Issue #8, opened 2022-05-17
Expose --require and --loader hook from this module.


DylanPiercey:
Would be great if the `require` and `loader` hooks from the other `esbuild-kit` modules were exposed from this top level one. In some scenarios I wan't to use the main `tsx` bin, but still the others are handy and it'd be nice if I didn't have to install other modules.

I would thinking it should be as simple as adding a `exports` to this module that with the `import` condition would expose the `loader` and otherwise would expose the require hook such that you could:

```terminal
node -r tsx
node --loader tsx
```

as an example of where this is useful:

```
mocha -r tsx
```

electrovir:
For any future visitors here: when switching from `ts-node/register`, use just `tsx` directly (`tsx/register` doesn't exist).

privatenumber:
It's also documented here:
https://github.com/esbuild-kit/tsx#nodejs-loader

electrovir:
Ah, nice! I was searching for `register`, which explains why I didn't find it in the docs :)

PR #9, opened 2022-05-17
feat: expose loader and require hook

DylanPiercey:
Implements #8

Exposes the underlying `--require` and `--loader` hooks from `@esbuild-kit/cjs-loader` and `@esbuild-kit/esm-loader` respectively.

Note: I think this is all good but I have not actually tested it yet.


DylanPiercey:
Looks like the cjs `--require` hook isn't working since the package is `type: "module"`. Need to figure out how to make pkgroll output `.cjs` files.

DylanPiercey:
I've updated the `exports` to expose `./loader` for the `--loader` and `./register` for the `--require` hook. This seems to work well (previously it was causing an issue in mocha specifically which tried to use esm first).

privatenumber:
Thanks for the PR @DylanPiercey 

I would actually prefer the original syntax you had so it can be used like:
```sh
node -r tsx --loader tsx
```

Can you try something like this?
```json5
{
	...,
	"exports": {
		".": {
			"require": "./dist/loader/cjs.cjs",
			"import": "./dist/loader/esm.mjs"
		}
	},
	...
}
```

You can test the change after pushing it to GitHub with: [build-this-branch](https://github.com/privatenumber/build-this-branch).

privatenumber:
Just pulled in the latest `develop` which makes a change to the export map.

DylanPiercey:
@privatenumber I did test with trying to have a single `.` export. However the problem is I'm personally trying to use this as a require hook with mocha which currently has some strange rules around supporting esm, like not supporting it in watch mode. The silly thing is that when you pass a `--require` flag to mocha it will always try to `import()` it first, and fallback to `require()`. Because of this I couldn't actually install the require hook and so I thought exposing them independently might be better.

Maybe there is a another option here though. I wonder if we could _only_ expose an esm version and have it do something like this:


```js
import { createRequire } from "module";
createRequire(import.meta.url)("@esbuild-kit/cjs-loader");

// export all of the loader hooks also:
export const resolve /** ... */
```

I don't have time to test this just yet.

DylanPiercey:
@privatenumber turns out I did have time.

I tested it and it seems to work great!
Essentially now when you import `tsx` from esm it will register both the loader and the require hook. If you import from cjs then it only adds the require hook.

DylanPiercey:
One thing to consider but I'm not sure if it should be included in this, is that we may want to update the following to use the internal imports rather than directly using `@esbuild-kit/*`:
https://github.com/esbuild-kit/tsx/blob/develop/src/run.ts#L38

src/loader/esm.ts
```diff
@@ -0,0 +1,3 @@
+import { createRequire } from "module";
+export * from "@esbuild-kit/esm-loader";
+createRequire(import.meta.url)("@esbuild-kit/cjs-loader");
```
privatenumber, line 3: FYI with pkgroll, you should be able to just do `require('@esbuild-kit/cjs-loader')`

privatenumber (commented):
Looks great!

Can you run the linter? `npm run lint -- --fix`

src/loader/esm.ts
DylanPiercey: The idea is that this is loaded from nodes esm environment, but still adds the cjs require hook so that if you manually createRequire it will still work. I don't think the above suggestion would work?

DylanPiercey: I guess it should be fine to import a cjs module which should also work. Never mind!

privatenumber: It will polyfill it:

https://github.com/privatenumber/pkgroll#require-in-esm

DylanPiercey: Should we even need that though? In theory this would just an import to a cjs module which _should_ just work right?

In theory I can just have

```js
export * from "@esbuild-kit/esm-loader";
import "@esbuild-kit/cjs-loader";
```

DylanPiercey:
Will be a few hours before I can update the commit. I'll double check that everything still works w/o using createRequire or require. If that doesn't work I'll just use require as you've said.

DylanPiercey:
@privatenumber I simplified as I described above and it still works. Also ran the lint/format. LMK if anything else is missing 😄

README.md
```diff
@@ -33,6 +33,24 @@ Pass in a file to run:
...
+Or pass `tsx` into the [`--require`](https://nodejs.org/api/cli.html#-r---require-module) flag if using cjs.
+```sh
+node -r tsx ./file.ts
+```
+
+This works well for tools that allow you to pass [Node.js CLI flags](https://nodejs.org/api/cli.html) but expose their own binary to run.
+For example for usage with [mocha](https://mochajs.org) you could run the following:
+```sh
+mocha -r tsx
+```
```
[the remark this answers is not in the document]
DylanPiercey: @privatenumber it does work if you use mocha in esm mode, and with our changes here it will work for cjs mode also. The only real problem is the warning that loaders are experimental which js kinda lame.

DylanPiercey:
@privatenumber I was just able to test this again, and actually I realized that with the current pattern of having the esm version both export the loader api, and register the require hook, it will still work as `mocha -r tsx` since the `-r` flag will work fine with the esm api.

I've gone ahead and simplified the PR now to only have a single `loader.js` and a simple `.` export.

privatenumber:
Awesome 👍 Glad we reached an elegant solution!

I removed the reference to mocha's `--require, -r` flag because they implement a custom one that [supports ESM](https://github.com/mochajs/mocha/blob/023f548213e571031b41cabbcb8bb20e458b2725/lib/nodejs/esm-utils.js#L33). This can be confused with Node.js's `--require, -r` flag, which [only supports CommonJS](https://nodejs.org/api/cli.html#:~:text=only%20commonjs%20modules%20are%20supported.%20).

DylanPiercey:
> This can be confused with Node.js's --require, -r flag, which [only supports CommonJS](https://nodejs.org/api/cli.html#:~:text=only%20commonjs%20modules%20are%20supported.%20).

Ah I assumed node did the same and allowed esm here. So that means if there is a case where you want to use this with cjs only then it wouldn't work. So maybe what was there was better?

privatenumber:
For CJS only use-cases, cjs-loader should be used directly.

tsx should only be used when requiring both ESM and CJS support.

src/loader.ts
```diff
@@ -0,0 +1,6 @@
+// Hook require() to transform to CJS
+require('@esbuild-kit/cjs-loader');
```
privatenumber, line 2: I changed this to `require()` from `import`.

When using `import`, the ESM loader ends up calling require after determining it's a CJS package. Calling `require()` directly short-circuits it.

privatenumber (approved):
I think this is ready to merge if this looks good to you @DylanPiercey

privatenumber:
Don't really follow what you're trying to say... but will move forward with this tomorrow morning.

BTW, I took a closer look at what mocha's doing. I recommend passing tsx in via env variable so that it can intercept both `require` and `import`. With `-r`, the loader API is ignored.

DylanPiercey:
@privatenumber since mocha has limited support for esm and so I was using it in cjs mode (eg watch mode does not work). For this case I actually do exclusively want the require hook.

However I think it works fine in cases where you just need the require hook to use the loader with what we've done here. Actually in mocha it works to use `mocha --loader tsx` even for cjs stuff, so I think that's the way to go.

PR #23, opened 2022-05-28
feat(watch): add `--clear-screen` flag

patheticGeek:
Related issue #25 

When running in watch mode, sometimes the user doesn't want the console to be cleared, so added this flag.
Plz accept.


privatenumber:
Can you open an issue and explain your use-case?

patheticGeek:
> Can you open an issue and explain your use-case?

Here, #25

src/watch.ts
```diff
@@ -42,7 +47,11 @@ export const watchCommand = command({
 			runProcess.kill();
 		}
 
-		process.stdout.write(clearScreen);
+		if (!options.noClearScreen) {
+			process.stdout.write(clearScreen);
+		} else {
+			console.log('🔄 tsx: restarting');
```
knpwrs, line 53: It might be cool to make `noClearScreen` a boolean or string to provide a custom message here.

privatenumber (changes requested):
Let's follow Vite's footsteps here: https://vitejs.dev/config/#clearscreen

Can you rename it to `clearScreen` and default it to `true`?

patheticGeek:
> Let's follow Vite's footsteps here: https://vitejs.dev/config/#clearscreen
> 
> Can you rename it to `clearScreen` and default it to `true`?

Yep will do it!

patheticGeek:
@privatenumber do you wanna change the message it logs?
Also gonna make it not log restarting on the first run cause thats start not a restart

ccorcos:
This flag doesn't appear in the `--help` documentation btw.

Issue #38, opened 2022-06-19
Cannot import named export from a typescript file in a workspace package if entry is using es modules and other package is not.


remorses:
Cannot import named export from a typescript file in a workspace package if entry is using es modules and other package is not.

tsx will throw

```
SyntaxError: The requested module 'ts-package/src/file' does not provide an export named 'x'
```

I know this could be considered a node limitation but maybe we can compile typescript files to esm if entry is esm (even if workspace package does not have `"type": "module"`

Repro: [https://github.com/remorses/tsx-repros](https://github.com/remorses/tsx-repros)

```
pnpm i
pnpm run import-ts-file
```

privatenumber:
I investigated this and it might be tough to fix.

When a ESM file imports a CommonJS file, you would imagine a loader could simply transform the CommonJS file to ESM syntax on import so it's interoperable.

However, Node.js actually [bypasses the CJS loader and reads the CommonJS file directly from disk](https://github.com/nodejs/node/blob/v18.4.0/lib/internal/modules/esm/translators.js#L211) to statically detect named exports with [`cjs-module-lexer`](https://github.com/nodejs/cjs-module-lexer).

Static analysis is likely necessary because the import statement must assert that the named export exists before running the file. And even if Node.js adds a hook to transform the file before statically analyzing it, we won't be able to support older Node.js versions.

I tackled this problem before (https://github.com/evanw/esbuild/issues/2248) but ran into the same conclusion. However, I managed to fix it for dynamic imports.

loynoir:
😢 Same issue here, go back to `esbuild-node-loader` for a while.

loynoir:
Found some `^bind.*$` and `^[a-z]{3}$` not working

privatenumber:
@loynoir 

`esbuild-node-loader` doesn't support this either so I think you're confusing this with a different issue.

> Found some `^bind.*$` and `^[a-z]{3}$` and `[A-Z]+_PROJECT_[A-Z]+` not working

I have no idea what you're talking about here. Can you open a new issue?

loynoir:
@privatenumber No, all three exports works.

loynoir:
```sh
$ tsx ./src/cli.mts
SyntaxError: The requested module './util' does not provide 
an export named 'XXX_PROJECT_XXX'

$ node --loader esbuild-node-loader ./src/cli.mts
OK
```

privatenumber:
Can you upload that reproduction?

[...]

PR #41, opened 2022-06-21
test: failing test for #38

privatenumber:


remorses:
I think this can be fixed by passing `platform: 'node'` to esbuild, this will emit cjs annotation like these
```js
0 && (module.exports = {
  x
});
```

Here is its documentation: https://esbuild.github.io/api/#platform

privatenumber:
I commented this on the issue, but I already went down that path via https://github.com/evanw/esbuild/issues/2248.

remorses:
Sorry but what you said in #2362 about directly reading from disk does not apply to #38, because in that case the cjs file is a typescript file, so it will be transformed anyway

privatenumber:
Do you have any research to back that up?

To me, it looks like a pretty straight forward `fs.readFileSync` without any consideration for the possibility of transformation: https://github.com/nodejs/node/blob/bde889bd4ed68d33163897b4175557fbc669b66f/lib/internal/modules/esm/translators.js#L211

PR #46, opened 2022-06-23
fix(repl): specify ts loader

vaakian:
# Problem
TypeScript REPL can't actually execute TypeScript expressions
# Reason
It detected a wrong loader(`js`) because in [core-utils](https://github.com/esbuild-kit/core-utils/blob/7348ce37265de7b76be726b9c4e6e335e65be6c0/src/transform/get-esbuild-options.ts#L35) that `path.extname('.ts')` returns an empty string, so it will add a `.js` suffix to the `filePath`and then a `js` loader is detected.

see [transform/get-esbuild-options.ts#L35](https://github.com/esbuild-kit/core-utils/blob/7348ce37265de7b76be726b9c4e6e335e65be6c0/src/transform/get-esbuild-options.ts#L35)

image
there's also another solution I prefer:

```js
transform(code, 'repl', { loader: 'ts' })
```

### DIFF
--- a/src/repl.ts
+++ b/src/repl.ts
@@ -14,7 +14,7 @@ const nodeRepl = repl.start();
 const { eval: defaultEval } = nodeRepl;
 
 const preEval: REPLEval = async function (code, context, filename, callback) {
-	const transformed = await transform(code, '.ts').catch(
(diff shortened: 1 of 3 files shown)


privatenumber:
Thanks for catching this & the PR!

PR #91, opened 2022-08-29
feat: add `package.json` subpath export

fz6m:
Like use :

```ts
  require.resolve('tsx/package.json')
```

Need export `package.json` subpath, otherwise an error will be thrown.

A real example : https://github.com/umijs/umi/pull/9152/files#diff-faa31e153df54d7263fcc895970fea96ab82253e8943d08d85ea22b49338ac0fR53

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
 		"dist"
 	],
 	"exports": {
+		"./package.json": "./package.json",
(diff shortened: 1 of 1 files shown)


privatenumber:
Thanks!

PR #111, opened 2022-10-05
feat: add '--test' flag

### LINKED ISSUE #85: Support new experimental built-in node test runner
### Feature request

NodeJS has a [new experimental built-in test runner](https://nodejs.org/docs/latest-v18.x/api/test.html) in version 18+. It is typically invoked for JavaScript files via `node --test` or `node --test test/**/*.test.js`, for example.

I would like to be able to invoke the new node test runner via `tsx`:

- `npx tsx --test` - ideal, but the test runner doesn't pick up `.ts` files by default
- `npx tsx --test test/**/*.test.ts` - explicit path(s) to test files

Currently when I try those commands, it does start the test runner and prints out the first line, then hangs and has to be interrupted. Looks like it is starting the node REPL, so maybe tweaking that behavior is all that's needed.

```bash
$ npx tsx --test test/**/*.test.ts
TAP version 13

# !!! Hangs here until I press Ctrl+C

# Subtest: /home/neodon/.local/share/pnpm/store/v3/tmp/dlx-34798/node_modules/.pnpm/tsx@3.8.2/node_modules/tsx/dist/repl.js
not ok 1 - /home/neodon/.local/share/pnpm/store/v3/tmp/dlx-34798/node_modules/.pnpm/tsx@3.8.2/node_modules/tsx/dist/repl.js
  ---
  duration_ms: 6.935629576
  failureType: 'cancelledByParent'
  error: 'test did not finish before its parent and was cancelled'
  code: 'ERR_TEST_FAILURE'
  stack: |-
    process.emit (node:events:513:28)
    process.emit (/home/neodon/.local/share/pnpm/store/v3/tmp/dlx-34798/node_modules/.pnpm/tsx@3.8.2/node_modules/tsx/dist/suppress-warnings.cjs:1:307)
  ...
# Subtest: /home/neodon/p/labs/test/foo.test.ts
ok 2 - /home/neodon/p/labs/test/foo.test.ts
  ---
  duration_ms: 0.082254723
  ...
1..2
# tests 2
# pass 1
# fail 0
# cancelled 1
# skipped 0
# todo 0
# duration_ms 6.964441232
```

### Why?

Currently I am able to run the node test runner with `tsx` by using it as a loader, so this feature is more of a minor thing / polish. It could be more intuitive to new users.

Example:

```typescript
// test/foo.test.ts
import test from 'node:test'
import assert from 'node:assert/strict'

test('node test!', async t => {
  assert.equal(true, true)
})
```

In `package.json`, I have `"type": "module"` set.

I can run the tests with:

```bash
$ node --loader tsx --test test/**/*.test.ts
TAP version 13
# Subtest: /home/james/p/labs/test/foo.test.ts
ok 1 - /home/james/p/labs/test/foo.test.ts
  ---
  duration_ms: 0.076968906
  ...
1..1
# tests 1
# pass 1
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 0.103634573
```

This is fine for me, but wouldn't it be awesome if you could run tests with `tsx --test`, analagous to `node --test`? 😎 

### Alternatives

_No response_

### Additional context

_No response_

romanlamsal:
closes #85 

After merging, `tsx --test` behaves like `node --test`. If no files are given, valid test files are searched based off the [specification](https://nodejs.org/api/test.html#test-runner-execution-model) but enhanced by including `.ts`, `.cts` and `.mts` files.

Also, glob patterns are working out of the box without passing a loader, so after this, instead of using
```sh
node --loader tsx --test **/*.test.ts
```
you can run
```sh
tsx --test **/*.test.ts
```


src/cli.ts
```diff
@@ -53,7 +59,9 @@ cli({
 		}
 
 		// Load REPL
-		process.argv.push(require.resolve('./repl'));
+		if (!argv.flags.test) {
+			process.argv.push(require.resolve('./repl'));
+		}
```
privatenumber, line 64: I would prefer an implementation where tsx just passes down `--test` to Node.js. In other words, `src/test-runner/util.ts` should not be added.

The reason it's not currently working is because tsx starts the REPL here if no arguments are passed in.

Instead, this condition should might be better as "if no arguments **or flags** are passed in, start the REPL". More specifically, we should check how Node.js behaves with each flag: do any of them complement the REPL or block it?

romanlamsal: Yes, I started out with exactly that - just prevent the REPL and pass down `--test whatever` to node.

Currently only `argv._` from the `typeFlag` call are passed down to node, effectively removing all flags. This behavior would need to be adjusted for a more generic approach, no? Something like "no flags, no matching command -> use `process.argv.slice(2)` instead".

Plus, the native `node --test` will not pick up typescript files unless a glob is provided. Just wanted to improve convenience here with the `test-runner/util`

romanlamsal: I'll gladly implement either approach 👍🏼  Just thinking about my own projects I'd rather have `tsx --test` pick up all typescript files as well

tests/index.ts
```diff
@@ -51,6 +51,10 @@ const nodeVersions = [
 				node.packageType = packageType;
 
 				await describe(`Node ${node.version}`, ({ runTestSuite }) => {
+					if (node.version >= '18') {
+						runTestSuite(import('./specs/test-runner'));
```
privatenumber, line 55: Did you check in all files? This file doesn't seem to exist.

tests/specs/cli.ts
```diff
@@ -54,5 +55,103 @@ export default testSuite(async ({ describe }, fixturePath: string) => {
...
+						path.join(fixturePath, 'log-argv.ts'),
+					],
+				});
+
+				// then
+				expect(tsxProcess.exitCode).toBe(0);
+				expect(tsxProcess.stdout).toMatch(logMessage);
+				expect(tsxProcess.stderr).toBe('');
+			});
+		});
```
privatenumber, line 82: This test seems to pass without your changes

tests/specs/cli.ts
```diff
@@ -54,5 +55,103 @@ export default testSuite(async ({ describe }, fixturePath: string) => {
...
+				});
+
+				// when
+				await tsxProcess;
+
+				// then
+				expect(aggregatedOutput).toMatch('tests 2');
+				await fixtures.rm();
+			});
+		});
```
privatenumber, line 155: Since tsx doesn't affect how Node.js test runner works, I'm not sure if these are relevant tests.

PR #115, opened 2022-10-07
fix(watch): relative paths in --ignore flag

romanlamsal:
When using relative paths with `watch --ignore`, chokidar won't pick up the correct files.

The existing test is refactored to use relative paths instead of absolute paths to ignore and execute. 

The fix is to explicitly provide a `cwd` to chokidar.

**Reproduction**: open [stackblitz](stackblitz.com/edit/node-q34cum?file=index.js&view=editor), run `pnpm dev` and change something in `ignored.js`. `tsx` should not rerun, but does.

### DIFF
--- a/tests/specs/watch.ts
+++ b/tests/specs/watch.ts
@@ -161,12 +161,13 @@ export default testSuite(async ({ describe }, fixturePath: string) => {
 				});
 
 				const tsxProcess = tsx({
+					cwd: fixture.path,
 					args: [
 						'watch',
 						'--clear-screen=false',
-						`--ignore=${path.join(fixture.path, fileA)}`,
-						`--ignore=${path.join(fixture.path, 'directory/*')}`,
-						path.join(fixture.path, entryFile),
+						`--ignore=${fileA}`,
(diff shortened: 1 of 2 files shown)


tests/specs/watch.ts
```diff
@@ -161,12 +161,13 @@ export default testSuite(async ({ describe }, fixturePath: string) => {
...
 				const tsxProcess = tsx({
+					cwd: fixture.path,
 					args: [
 						'watch',
 						'--clear-screen=false',
-						`--ignore=${path.join(fixture.path, fileA)}`,
-						`--ignore=${path.join(fixture.path, 'directory/*')}`,
-						path.join(fixture.path, entryFile),
+						`--ignore=${fileA}`,
+						'--ignore=directory/*',
```
privatenumber, line 169:
```suggestion
						`--ignore=${path.join(fixture.path, 'directory/*')}`,
```

Let's keep one absolute paths so we can test both.

privatenumber:
Thanks for the fix @romanlamsal

PR #133, opened 2022-11-01
fix: follow Node.js flag handling

privatenumber:
- `tsx --version` now shows tsx's version, and passes the flag down to Node.js to show the version too
- `tsx --help` now shows tsx's help, and passes the flag down to Node.js to show the help too
- No longer implicitly accepts values to unknown flags (fix https://github.com/esbuild-kit/tsx/issues/70)


src/cli.ts
```diff
@@ -15,54 +18,44 @@ const tsxFlags = {
...
-		process.argv.slice(2),
-		{ ignoreUnknown: true },
-	)._;
-
-	const childProcess = run(args, {
-		noCache: Boolean(argv.flags.noCache),
-		tsconfigPath: argv.flags.tsconfig,
-	});
+	const childProcess = run(
+		removeArgvFlags(tsxFlags),
```
fz6m, line 53: This is a breaking change.

When pass args to `some.ts` :

**< 3.12.0** :

```bash
  tsx some.ts -- dev
```

**>= 3.12.0** :

```bash
  tsx some.ts dev
```

follow [Semantic Versioning Specification](https://github.com/semver/semver) , I think should bump to `4.0.0` , not `3.12.0`

PR #157, opened 2022-12-04
chore: add `.editorconfig`

IlyaSemenov:
I've noticed that (almost) entire repo uses tabs which is not in sync with the default settings of most IDEs. This PR adds `.editorconfig` which expresses the coding style explicitly. It also fixes a few residual scripts which were using spaces.

Also, `tests/specs/cli.ts` was failing to pass the pre-commit hook due to an unused variable, I included the fix in this commit as well.

### DIFF
--- a/tests/specs/cli.ts
+++ b/tests/specs/cli.ts
@@ -5,7 +5,7 @@ import { tsx, tsxPath } from '../utils/tsx';
 import { ptyShell, isWindows } from '../utils/pty-shell';
 
 export default testSuite(({ describe }, fixturePath: string) => {
-	describe('CLI', ({ describe, test }) => {
+	describe('CLI', ({ describe }) => {
 		describe('version', ({ test }) => {
 			test('shows version', async () => {
 				const tsxProcess = await tsx({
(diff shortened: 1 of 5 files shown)


tests/specs/cli.ts
```diff
@@ -5,7 +5,7 @@ import { tsx, tsxPath } from '../utils/tsx';
 import { ptyShell, isWindows } from '../utils/pty-shell';
 
 export default testSuite(({ describe }, fixturePath: string) => {
-	describe('CLI', ({ describe, test }) => {
```
privatenumber, line 8: Why's this removed? It's used on https://github.com/IlyaSemenov/tsx/blob/8b3ec3939876c3dd0077854f2fa9e74378f57654/tests/specs/cli.ts#L62

privatenumber:
`tests/fixtures` is ignored by ESLint so you shouldn't be getting linting errors there

tests/specs/cli.ts
IlyaSemenov: My bad. I disabled that tests as it was failing on my machine, and didn't realise _that_ in turn was causing the linter error above.

PR #158, opened 2022-12-04
fix invalid tsconfig.extends under tests

IlyaSemenov:
`tsconfig.extends` must point to the base `tsconfig.json`, [possibly using Node.js style resolution](https://www.typescriptlang.org/tsconfig#extends). Pointing to a directory is non-standard way that may work in certain environments but then it stops. (I actually used to do the same earlier until I was hit by the same problem when upgrading the stack.)

For instance, it's not currently possible to build the tests with plain `esbuild`:

```
~/gh/tsx/tests/fixtures/tsconfig/node_modules/should-not-resolve-paths fix/tsconfig_extends*
❯ /Users/semenov/gh/tsx/node_modules/.pnpm/esbuild@0.15.10/node_modules/esbuild/bin/esbuild index.mjs
✘ [ERROR] Cannot read file "../../../../..": is a directory

    ../../../../tsconfig.json:2:12:
      2 │   "extends": "..",
        ╵              ~~~~

1 error
```

It works after the patch:

```
~/gh/tsx/tests/fixtures/tsconfig/node_modules/should-not-resolve-paths fix/tsconfig_extends*
❯ /Users/semenov/gh/tsx/node_modules/.pnpm/esbuild@0.15.10/node_modules/esbuild/bin/esbuild index.mjs
import value from "p/nested-resolve-target";
console.log(value);
```

### DIFF
--- a/tests/tsconfig.json
+++ b/tests/tsconfig.json
@@ -1,4 +1,4 @@
 {
-	"extends": "..",
+	"extends": "../tsconfig",
 	"include": ["."]
 }


privatenumber:
This is a TypeScript spec, so you'll have to demonstrate that it's not working with TypeScript rather than esbuild.

As for esbuild not being able to resolve `..`, you should file a bug with them instead.

IlyaSemenov:
@privatenumber It's not a typescript spec, I specifically linked the spec above.

The spec says: _"The value of extends is a string which contains **a path to another configuration file** to inherit from. The path may use Node.js style resolution."_

The spec doesn't say that the string contains a path to a directory containing another configuration file, it must contain a path to the file itself (possibly without extension).

privatenumber:
Can you demonstrate that it's not working using TypeScript?

IlyaSemenov:
@privatenumber See https://stackblitz.com/edit/node-aiwgcb?file=tsconfig.json

With this `tsconfig.json`:

```json
{
  "extends": "./foo",
  "compilerOptions": {
    "noEmit": true
  }
}
```

it says:

```
❯ tsc -p .
tsconfig.json:2:14 - error TS6053: File './foo' not found.

2   "extends": "./foo",
               ~~~~~~~


Found 1 error.
```

If you change `"./foo"` to `"./foo/tsconfig"` it will compile normally.

The Typescript specification is very clear about it.

IlyaSemenov:
Besides, your other tsconfig does it correctly: https://github.com/esbuild-kit/tsx/blob/e77e9c1c86b88f5988d2ceff815864824bfa3e45/tests/fixtures/tsconfig/tsconfig-custom/tsconfig.custom-name.json#L2

So at the very least it's inconsistent across the project.

privatenumber:
Your reproduction is not the same.

Swap the `extends` around so `foo/tsconfig.json` is extending `..`, `cd` into `foo`, and run `tsc --showConfig`.
You'll see there's no problem here.

Feel free to file a bug with esbuild to follow TypeScript behavior in `extends`.

IlyaSemenov:
OK then. I see this "fixed" in esbuild 0.15.18 (still "broken" in 0.15.17, and also in 0.15.10 which is pulled by `@esbuild-kit/*`).

I filed an upstream problem to Typescript as this behaviour is inconsistent and against the specs (both the tsconfig specification and Typescript 3.2 changelog which introduced bare path extends): https://github.com/microsoft/TypeScript/issues/51782

privatenumber:
> OK then. I see this "fixed" in esbuild 0.15.18 (still "broken" in 0.15.17, and also in 0.15.10 which is pulled by `@esbuild-kit/*`).
 
tsx doesn't use `esbuild` to resolve `tsconfig.json`. What's the relevance of this?

PR #190, opened 2023-02-28
feat: support TypeScript code in eval flag

### LINKED ISSUE #162: --eval support
### Feature request

Ability to use typescript in using the node `--eval` or `-e` feature from node:

- https://nodejs.org/api/cli.html#-e---eval-script.

eg api:

```
tsx  -e "const thing: string = 'hi'; console.log(thing);"
```

Currently this shows the loader is not applied with the error:

```
[eval]:1
const thing: string = hi; console.log(thing);
      ^^^^^

SyntaxError: Missing initializer in const declaration
    at new Script (node:vm:100:7)
    at createScript (node:vm:257:10)
    at Object.runInThisContext (node:vm:305:10)
    at node:internal/process/execution:81:19
    at [eval]-wrapper:6:22
    at evalScript (node:internal/process/execution:80:60)
    at node:internal/main/eval_string:27:3
``` 

### Why?

This is a feature used commonly in the node ecosystem for example code and running code snippets.

It can be also be useful for running a script without having to create a file. Right now creating a file is the only way eg related question here about [Programmatic usage](https://github.com/esbuild-kit/tsx/discussions/102#top).

### Alternatives

_No response_

### Additional context

Initial question here https://github.com/esbuild-kit/tsx/discussions/161 where @privatenumber suggests this should be possible.

rayrw:
fixes #162 

This PR is my attempt to add support of TypeScript code in the eval flag.
Not sure if it's the correct approach. I've got inspired by the `preEval` function of `repl.ts`, so I tried doing something similar by using `transformSync` from `@esbuild-kit/core-utils`. After transforming it, I tried swapping the original TS code's argv with the transformed code.

There are still several things I'm not sure (e.g. the `filePath` parameter of `transformSync`).
Please guide me if I've done it wrong.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,4 +1,5 @@
 import { cli } from 'cleye';
+import { transformSync } from '@esbuild-kit/core-utils';
 import { version } from '../package.json';
 import { run } from './run';
 import { watchCommand } from './watch';
@@ -36,6 +37,11 @@ cli({
 			alias: 'h',
 			description: 'Show help',
 		},
+		eval: {
+			type: String,
+			alias: 'e',
+			description: 'Evaluate code inside the eval flag',
+		},
 	},
 	help: false,
 	ignoreArgv: ignoreAfterArgument(),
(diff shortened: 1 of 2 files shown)


privatenumber:
Thanks, I haven't played with it yet, but the code looks good.

Can you add a test with the [`--input-type`](https://nodejs.org/api/cli.html#--input-typetype) flag to make sure `module` works expectedly?

Off the top of my head, maybe something like `console.log(import.meta.url)` or making sure this fails `console.log(__dirname)`.

rayrw:
@privatenumber Thanks for the quick replies!
Yeah, testing the `--input-type` flag sounds very useful.
I've compared the output between normal `node` cli and `tsx`, and added the test cases.
Hopefully it's the expected behavior (eval `import.meta.url` returns `undefined` under `--input-type=module` flag).

PR #194, opened 2023-03-08
fix: hide relaySignal from `process.listeners()`

### LINKED ISSUE #193: SIGTERM does not cause exit when signal-exit is used
### Bug description

If you run a program with tsx which uses signal-exit (for running some code before exit) then CTRL-C will not quit the program. In my case I was trying to use [Ink](https://github.com/vadimdemedes/ink) inside tsx (ink depends on hide-cursor which depends on signal-exit). This issue is caused because both tsx and signal-exit don't want to quit the process because from each of their perspectives there is another listener which would normally prevent the app from quitting.

Related: #122 

### Reproduction

https://github.com/jakzo/tsx-signal-exit-issue

Run `npm run start:node` to see expected behaviour.

### Environment

```shell
System:
    OS: macOS 13.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 86.52 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.15.1 - ~/.nvm/versions/node/v16.15.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.15.1/bin/yarn
    npm: 8.11.0 - ~/.nvm/versions/node/v16.15.1/bin/npm
  npmPackages:
    tsx: ^3.12.3 => 3.12.3
```

### Can you work on a fix?

- [X] I’m interested in opening a pull request to address this issue.

jakzo:
Closes #193 

Continues the approach from #123 and hides the `relaySignal` listener added by tsx from `process.listeners()` which is what the signal-exit module uses to determine whether to proceed with exiting the program. I've tested and confirmed that it fixes the issue.

https://github.com/tapjs/signal-exit/blob/8d4bd9c/index.js#L121

### DIFF
--- a/src/preflight.cts
+++ b/src/preflight.cts
@@ -45,4 +45,14 @@ if (process.send) {
 		}
 		return count;
 	};
+
+	// Also hide relaySignal from process.listeners()
+	const { listeners } = process;
+	process.listeners = function (eventName) {
+		const result = Reflect.apply(listeners, this, arguments);
+		if (relaySignals.includes(eventName as any)) {
+			return result.filter((listener: any) => listener !== relaySignal);
+		}
+		return result;
+	};
 }


privatenumber:
Thanks for the well documented issue and fix @jakzo !

PR #220, opened 2023-04-10
Add flag to control restart-on-enter behavior

### LINKED ISSUE #215: Support for flag to disable clear screen on return
### Feature request

Thanks for developing and maintaining this library. Would you be open to adding a flag to disable the restart on return behavior when running in watch mode ? 

### Why?

When server generates a lot of logs, I often use returns to create a visual gap between  past output and next output before performing some action so that it is easier to see eg. what all db queries were executed as part of a specific request handling. 

I don't mind manually restarting through `Ctrl+c ctrl+p enter`  but every now and then I find myself accidentally restarting server when trying to add some vertical space in console output out of muscle memory. 
 

### Additional context

_No response_

nirajsavaika:
Resolves #215

### DIFF
--- a/src/watch/index.ts
+++ b/src/watch/index.ts
@@ -35,6 +35,11 @@ const flags = {
 		type: [String],
 		description: 'Paths & globs to exclude from being watched',
 	},
+	restartOnReturn: {
+		type: Boolean,
+		description: 'Restart on pressing return key',
+		default: true
+	}
 } as const;
 
 export const watchCommand = command({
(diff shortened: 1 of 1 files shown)


atesgoral:
@privatenumber having this flag would be very valuable. We were unable to use tsx to run a console application in watch mode. 🙏

onemen:
Are you going to merge this PR?

privatenumber:
Likey won't be adding a new flag, but rather change the manual reload to a unique shortcut (e.g maybe Ctrl+R?).

privatenumber:
Closing as I'll be adding a different solution

PR #244, opened 2023-05-18
fix(watch): respawn process only after exit

npdev453:
That code cause EADDRINUSE error in 50% of restarts in watch command if port binding going fast. 
Aslo impossible to get around it with graceful shutdown logic.


src/watch/index.ts
```diff
@@ -98,6 +86,25 @@ export const watchCommand = command({
...
 		});
+	};
+
+	const reRun = debounce(() => {
+		if (
+			runProcess
+			&& (!runProcess.killed && runProcess.exitCode === null)
+		) {
+			runProcess.kill();
+		}
```
privatenumber, line 97: Can you move this logic down right after the `exit` handler is registered?

npdev453: @privatenumber like this?

ben-xD: @privatenumber, could you take a look? Fixing this problem would improve developer experience quite a lot

giva9712: it is a very useful feature, PTAL @privatenumber

privatenumber:
Maintainer editing is disabled on this pull request, so I can't make changes or run tests.

Closing in favor of https://github.com/esbuild-kit/tsx/pull/298

Issue #253, opened 2023-06-10
SIGINT handler doesn't wait for child proc to close, but instead kills it immediately


NinjaInShade:
#### What did you do? (Provide reproduction code in next section)
  
  I have a SIGINT handler that does some disposing and when running with TSX in watch mode, and SIGINT is received, it never waits for the child process to actually close, it just kills it immediately.

  #### What did you expect to happen?
  
  To let SIGINT handler function to fully run and exit if event loop empty or if the handler function calls process.exit().

  #### What happened instead?
  
  It doesn't wait, it just kills the child process immediately, so dispose code never runs fully.

If I comment out line 130 in `src/watch/index.ts` *some*/*more* dispose code runs, but definitely isn't waiting for event loop to finish or process.exit() to be called

  #### Do you have an error stack-trace or screenshots?

  Nothing to report in terms of "errors" here.

### Reproduction

You can see this with this example code:
```
import http from 'node:http';

const server = http.createServer();
server.listen(5605);

process.on('SIGINT', async () => {
    console.log('Got SIGINT');

   setTimeout(() => {
      console.log('finished');
   }, 3000);
});
```

And then run this file with `tsx file.ts` and trigger a SIGINT through typing `Ctrl-C` in the terminal.

If you ran this with just `node` - provided you change it to plain javascript, no typescript extension - it would wait to dispose properly

### Environment

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 AMD Ryzen 7 7700X 8-Core Processor
    Memory: 1018.85 MB / 31.15 GB
  Binaries:
    Node: 19.8.0 - C:\Program Files\nodejs\node.EXE
    npm: 9.6.5 - C:\Program Files\nodejs\npm.CMD
  npmPackages:
    tsx: ^3.12.7 => 3.12.7
```

### Can you work on a fix?

- [ ] I’m interested in opening a pull request to address this issue.

midzelis:
My latest commit [7467d2f](https://github.com/esbuild-kit/tsx/pull/254/commits/7467d2f264d8dfac08e48b67b728f9b6e2728162) improves/fixes your use case. 

I also needed to modify your example like so

```js
import http from 'node:http';

const server = http.createServer();
server.listen(5605);

process.on('SIGINT', async () => {
    console.log('Got SIGINT');

   setTimeout(() => {
      console.log('finished');
      server.close();
      process.removeAllListeners()
   }, 3000);
});
[...]

Issue #252, opened 2023-06-09
ctrl-c (SIGINT) when launching via --inspect-brk and using dedicated DevTools for Node will not term process or debugger


midzelis:
### Bug description

Open Chrome Dedicated DevTools for Node
Run tsx --debug-inspect
Press ctrl-c, this does not terminate the process. 

If you run this via node directly, it will work as expected: Pressing ctrl-c will stop the program and disconnect it from Chrome DevTools. 

### Reproduction

1. Create a simple hello world app with a debugger
2. Open chrome://inspect and click "Open dedicated DevTools for Node" 
3. Run `tsx --debug-inspect hello.ts`
4. Chrome will auto start debugging
5. Press ctrl-c  - process will not terminate, but chrome debugger not active anymore

### Environment

```shell
tsx 3.12.7, chrome Version 114.0.5735.106 (Official Build) (x86_64), node v18.16.0
```

### Can you work on a fix?

- [ ] I’m interested in opening a pull request to address this issue.

privatenumber:
Just looked into this

1. `--debug-inspect` is not a valid flag.

2. Tried with `--inspect` and a file that contains `setTimeout(() => {}, 1e5)` and I could <kbd>Ctrl+C</kbd> just fine.

Closing as I can't reproduce this (maybe it was fixed recently? but I can't verify because not sure how to repro). Happy to reopen if you can provide a reproduction.

Issue #201, opened 2023-03-16
When launched with tsx, process.on('message') and process.send doesn't work


FishingHacks:
### Bug description

I launched tsx with my ts-file, using `child_process` and with ipc attached (`spawn('tsx', ["<filepath>"], {stdio: ["ipc"]})`).
I then listened for messages, and sent a message to the process, which the process should respond to, and then exit, but instead, nothing happened. It doesn't throw an error either. I hope this description is clear enough.

### Reproduction

https://github.com/FishingHacks/tsx-error-ipc

Requirements:
- Globally installed tsx
- Globally installed ts-node (it will launch `b.ts` with ts-node as a comparison, that code-part is not neccessary)

### Environment

```shell
System:
    OS: Linux 5.4 Linux Mint 20.3 (Una)
    CPU: (4) x64 Intel(R) Core(TM) i5-3450 CPU @ 3.10GHz
    Memory: 11.36 GB / 15.57 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.6.0 - /usr/bin/node
    npm: 9.4.0 - /usr/bin/npm
```

### Can you work on a fix?

- [ ] I’m interested in opening a pull request to address this issue.

FishingHacks:
it appears that 
```typescript
if (process.send) {
    let handledDep = false;
    childProcess.on('message', (message: any) => {
        if (message && message.type === 'dependency') handledDep = true;
        else if (handledDep && (!message || message.type !== 'kill'))
            process.send?.(message);
    });
    process.on('message', (msg: any) => childProcess.send(msg));
}
```
after the relaySignal function definition, it essentially fixes the error. Then i noticed, that `process.off` does not work, but `process.removeListener` does, so i added `process.off = process.removeListener;` at the end of `preflight.cts`.

When i tried to run `npm test`, it "froze" at `✔ Package type: commonjs › tsx › watch › ignore › file path & glob`, but that does not appear to be caused by the changes i made, as reverting them also displayed the same behavior.

privatenumber:
tsx is a wrapper process, and I'm not sure if I recall relaying IPC from the actual Node.js process to the parent. I think this will have to be researched & implemented.

I'm not sure if I understand the point of your last message—what are you trying to do?

alanhamlett:
Looks like an attempt was made to fix this in #206, but it's been a while and that PR seems stalled.

PR #254, opened 2023-06-10
Fix signal handling

midzelis:
Fixing my own reported issue #252 
Though, this might also for for #253 and issue #201

The problem is that the previous signal handling didn't work properly. Previously, on a SIGINT/SIGTERM signal, the parent would install a message listener, and wait for the childProcess to send a 'kill' IPC message to the parent (and then, it would also exit the process, if there were no other signal listeners installed). The problem is that node will not terminate a process if any open IPC listeners. 

Also, this simplifies the logic a bit. AFIACT, when a signal is sent to a process (using kill), it is sent to all processes of that process group. When a signal is sent via the terminal, it is sent to the child process. 

This commit simplifies the logic such that all signals from the parent are always sent to the child, but the parent also makes a note of which signal this is (for use later.) 

The child does not need to have any special signal handling. By default, SIGINT will return 128+2, and SIGTERM will return 128+15. If a child process chooses to implement their own signal handler - thats fine. That child process can have its own return code that isn't based on signal. If thats the case, it is used. Otherwise, the default exit code will be the 128+signalCode. 

This also supports SIGQUIT of the parent. By default, kill -9 of the parent will just kill the parent, but not the child. This leaves a dandling/orphaned process. To fix this, the clientProcess installs a on('disconnect') listener which will exit the child process when the parent goes away. (Note, this can only be SIGQUIT, since SIGINT/SIGTERM would be relayed to the child, SIGQUIT is the only one that can not be relayed) However, adding 'disconnect' to process will cause the child to prevent the process from existing. So a special `process.channel.unref();` is used to avoid this "leaked listener" from preventing the process to terminate. (https://nodejs.org/api/process.html#processchannelunref) 

This is probably ok. But just calling out a specific gotcha that may affect code that uses their own IPC handler for their own spawned children. preflight is basically imposing that IPC channels will not prevent the process from exiting, which may be unexpected. 

Technically, the entire lines 17-29 in preflight are all optional. They are only used to clean up a zombie process. But since it may cause unexpected behavior in some scripts that use their own IPC, this might need to have a flag. Or maybe just removed completely.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+    "singleQuote": true,
+    "editorconfig": true,
+    "trailingComma": "all",
+    "bracketSameLine": true
+}
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,76 +1,13 @@
-import { constants as osConstants } from 'os';
 import { cli } from 'cleye';
 import { version } from '../package.json';
 import { run } from './run';
 import { watchCommand } from './watch';
 import { removeArgvFlags, ignoreAfterArgument } from './remove-argv-flags';
-import { ChildProcess, type Serializable } from 'child_process';
-
-type SignalMessage = Serializable & {
-	type: 'signal';
-	signal: NodeJS.Signals;
-};
-
-// The interval between checks to see if child has acknowledged a signal.
-// Practically speaking, this is the minimum amount to wait before a signal
-// that is sent only to parent will be relayed to child.
-const POLL_INTERVAL_MS = 100;
-// The maximum time to wait to see if child has acknowledged a signal.
-// Practically speaking, if the client is in a tight synchronous loop, or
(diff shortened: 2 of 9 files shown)


midzelis:
So, just realized that this will cause the signals to be sent to the childProcess twice. Because, shells, by default, send the signal to the process group. Whereas 'kill' will send it just to the process specified. You can make `kill` send it to all members of the process group by using the syntax `kill -2 -GROUP_ID` where GROUP_ID is the parent process id/process group id. To eliminate the double signal, a solution similar what exists before this commit could be done - which effectively has the child inform the parent of the signal. The parent could then be modified to have a small window of time where its listening for multiple notifications, and to avoid relaying if it notices a response from the client. This may cause the client to prevent the process from existing. I'm going to investigate. 

Of course, that 'small window of time' is a bit arbitrary. If you are, for example, in a debugger, the client won't respond quickly, and will get double notified anyways.

PR #298, opened 2023-08-31
fix(watch): wait for the process to exit

privatenumber:
fix https://github.com/esbuild-kit/tsx/issues/82

- When hitting Enter
  - If the child process is not running, it simply re-runs it
  - If the child process is running, it kills it and waits for it to exit
    - If Enter is hit again while the child process is exiting, it force-kills it


stephenh:
Awesome! Thanks @privatenumber ! Will check this out later today...

PR #307, opened 2023-09-12
feat: If tsconfig is provided, watch it!

Dunqing:
The tsconfig may change, we need to be aware of its changes and re-run the


michaelangeloio:
+1

PR #333, opened 2023-10-05
fix(watch): listen to stdin only when `isTTY` and check for `\n` character

aleclarson:


src/watch/index.ts
```diff
@@ -195,5 +195,11 @@ export const watchCommand = command({
 	).on('all', reRun);
 
 	// On "Return" key
-	process.stdin.on('data', reRun);
+	if (process.stdin.isTTY) {
+		process.stdin.on('data', (input) => {
+			if (input.toString() === '\n') {
+				reRun();
+			}
```
privatenumber, line 202: Was checking for `\n` actually necessary for something? Or is it just for the sake of explicitness?

PR #350, opened 2023-10-20
Upgrade to esbuild 0.19

nwalters512:
This PR upgrades to the latest version of esbuild.

There are now two copies of esbuild installed in development, aas `pkgroll` is still using esbuild 0.18. I'd be happy to file a PR against that repository as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
 		"*.{js,ts,mjs,mts,cjs,cts,json}": "pnpm lint"
 	},
 	"dependencies": {
-		"esbuild": "~0.18.20",
+		"esbuild": "~0.19.5",
 		"get-tsconfig": "^4.7.2",
 		"source-map-support": "^0.5.21"
 	},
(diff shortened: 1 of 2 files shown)


privatenumber:
Thanks but we had to revert due to a change in how esbuild compiles `require()`:

https://github.com/esbuild-kit/core-utils/issues/47

Closing this for now.

nwalters512:
Thanks for linking to that other issue. I did search in this repository and check the Git blame before opening this PR, but the fact that it was in a different repository explains why I didn't see it!

LukeNotable:
Well, I had the same problem with `sharp` as Mariuz, and I just tried his [repo](https://github.com/esbuild-kit/core-utils/issues/47#issuecomment-1695442447), using the latest `tsx` with a resolution for `esbuild@0.19.8`, and suddenly at least that issue with `sharp` has gone away. Strange, since the bug in `esbuild` is still there. Did `tsx` accidentally implement a workaround somewhere along the way?

PR #355, opened 2023-10-26
fix(cache): use `process.geteuid()` for virtual users

ST-DDT:
Fixes https://github.com/esbuild-kit/tsx/issues/352

- https://github.com/esbuild-kit/tsx/issues/352
- https://github.com/esbuild-kit/tsx/pull/332


src/utils/transform/cache.ts
```diff
@@ -8,6 +8,20 @@ const getTime = () => Math.floor(Date.now() / 1e8);
...
 const tmpdir = os.tmpdir();
 const noop = () => {};
+
+/**
+ * Cache directory name is based on the user's uid.
+ * This is to avoid permission issues when as a different user in parallel.
+ */
+function cacheDirectoryName(): string {
+	const { geteuid } = process;
+	if (geteuid === undefined) {
```
privatenumber, line 18: I realized this might not exist on Windows.

I'm currently refactoring the tests to be lighter so I can bring back Windows. Will hold this until then.

privatenumber:
Windows doesn't seem to have a user ID so I'm considering using [node-cachedir](https://github.com/LinusU/node-cachedir) to store cache based in the user's home directory:

dcousens:
What about `os.userInfo().username`?

privatenumber:
What's wrong with using [node-cachedir](https://github.com/LinusU/node-cachedir)?

dcousens:
https://github.com/LinusU/node-cachedir/blob/1d8619a927cfc6cafb4543ed3ce46bc6558be99f/index.js#L5

```
sudo -u nobody node -e "console.log(path.join(os.homedir(), '.cache'))"
/.cache
```

That isn't going to help in your CI environments

dcousens:
I think 2051f3e665f091d231a5d8c759d849e86ba87089 as written did a pretty good job, is cache sharing a problem on Windows?

ST-DDT:
How about using `"tsx-" + Math.random().toString().substring(2)` => `tsx-354313983399`? 

Works on all platforms and collisions are very unlikely. Or is there an implicit need to reuse the folder for the same user?

dcousens:
> Or is there an implicit need to reuse the folder for the same user?
Yes, your proposal is equivalent to `tsx --no-cache`, the cache is shared between runs (AFAIK)

privatenumber:
On Linux, each user needs its own cache directory so it has the correct permissions to write to it. I'm not sure if this is a problem on Windows, but I rather keep the behavior consistent and also because the organization makes sense.

Do yall know if there's a way for me to test in the Linux virtual user environment? Hard to address it when I can't test the changes.

> How about using `"tsx-" + Math.random().toString().substring(2)` => `tsx-9675700890192862`? 

This defats the point of the cache. (Module caching already happens inside Node). The cache is shared across multiple runs per user so it needs to be deterministic.

ST-DDT:
I think on windows you always have a distinct username.

So Posix -> uid, Windows/Other -> username

ST-DDT:
That way

> Do yall know if there's a way for me to test in the Linux virtual user environment? Hard to address it when I can't test the changes.

You could use a docker container for that.

ST-DDT:
@privatenumber Did you close this PR by accident?

privatenumber:
Oops, looks like it closed as a side-effect of auto-deletion happening on a merged PR.

The reopen button is disabled for me—I think because it's your PR. Would you mind re-opening it?

ST-DDT:
I cannot because you closed it.
Maybe you have to change the target branch to an existing one before reopening is possible.
Otherwise I can open a new PR tomorrow.

PR #377, opened 2023-11-10
perf: exclude source maps for `node_modules`

privatenumber:


privatenumber:
Closing because:
- This will be a breaking change
- There's no way for users to enable source-maps in dependencies that distribute it
- Source maps will be disabled by default in next major so performance will be better by default and it's user toggleable https://github.com/privatenumber/tsx/discussions/386

PR #390, opened 2023-11-15
perf(esbuild): disable `sourcesContent`

privatenumber:
This disables [sourcesContent](https://esbuild.github.io/api/#sources-content) in esbuild:

> The sourcesContent field is an optional field that contains all of the original source code. This is helpful for debugging because it means the original source code will be available in the debugger.
>
> However, it's not needed in some scenarios. For example, if you are just using source maps in production to generate stack traces that contain the original file name, you don't need the original source code because there is no debugger involved. In that case it can be desirable to omit the sourcesContent field to make the source map smaller:

This doesn't impact the tests because it only checks the error stack. But I think it might make sense to investigate how this impacts the debugger first in different environments, and investigate whether there's a way to detect whether the debugger is enabled.


privatenumber:
:tada: This issue has been resolved in v4.3.0

PR #392, opened 2023-11-16
fix(source map): handle source map in middle of file

### LINKED ISSUE #389: Source map stripping breaks with source map comments in middle of file
### Precheck

- [X] I searched existing issues before opening this one to avoid duplicates
- [X] I'm able to reproduce this issue and prove it with a minimal reproduction
- [X] I understand this is not a place to ask for free debugging support

### Problem

The source map stripping added in #374 can change the behavior of files if the source map comment happens to be in the middle of the file. For instance:

```
type Name = string;

function sayHello(name: string) {
  console.log('Hello', name);
}

//# sourceMappingURL=index.js.map

sayHello('World');
```

For this code, `Hello World` will not be printed.

For better or worse, this kind of thing shows up in real-world code: https://github.com/drudru/ansi_up/blob/a30bb48417c9d67d8743ea30937b3b432b7648ec/ansi_up.js#L448. You can see how stripping the final lines along with the source map comment will result in unrunnable code that fails to parse.

### Expected behavior

I would expect source map stripping to not change the runtime behavior of files, even if the file happens to be malformed. Note that I am not trying to make the argument that this comment should be interpreted as a source map or otherwise handled by `tsx`.

I'd be happy to open a PR for this if we can decide on what the behavior should be. I'm thinking that we should only strip this comment if it's the last line of the file.

### Minimal reproduction URL

https://github.com/nwalters512/tsx-source-maps-parse-error-repro

### Version

v4.1.2

### Node.js version

v20.7.0

### Package manager

npm

### Operating system

macOS

### Contributions

- [X] I plan to open a pull request for this issue
- [ ] I plan to make a financial contribution to this project

nwalters512:
Fixes #389.


privatenumber:
Thanks for the PR and issue!

I had to refactor the regex approach because previous use of regex has proven to be not as performant on larger files

privatenumber:
:tada: This issue has been resolved in v4.1.3

PR #395, opened 2023-11-17
fix: send SIGKILL on unresponsive process

### LINKED ISSUE #394: SIGINT has no effect if tsx is running an infinite loop
### Precheck

- [X] I searched existing issues before opening this one to avoid duplicates
- [X] I'm able to reproduce this issue and prove it with a minimal reproduction
- [X] I understand this is not a place to ask for free debugging support

### Problem

You have a file called `file.ts` with an infinite loop in it like `while(true){}`.
By running `tsx file.ts` you won't be able to stop the execution of tsx by sending a SIGINT (for example with CTRL+C) and the process will run indefinitely until it receives another signal like a SIGQUIT or a SIGSTOP.

### Expected behavior

tsx should stop the execution after the SIGINT.

### Minimal reproduction URL

https://stackblitz.com/edit/node-fc7ukf?file=index.ts

### Version

v4.1.3

### Node.js version

v20.7.0

### Package manager

pnpm

### Operating system

macOS

### Contributions

- [x] I plan to open a pull request for this issue
- [ ] I plan to make a financial contribution to this project

sadkebab:
Fixes #394


src/cli.ts
```diff
@@ -86,6 +86,13 @@ cli({
 		 */
 		if (!message) {
 			childProcess.kill(signal);
+			/**
+			 * if the child process is stuck in a loop, SIGINT
+			 * and SIGTERM have no effect on it, therefore we send
+			 * a SIGKILL if the childProcess.on('close') handler
+			 * defined below is not called within 10ms.
+			 */
+			setTimeout(() => childProcess.kill('SIGKILL'), 10);
```
privatenumber, line 95: This will break cases where the child handles the `SIGINT` signal, either by refusing to close or running a slow handler:
```
process.on('SIGINT', () => {
	// slow code
});
```

PR #405, opened 2023-11-22
refactor: remove source-map-support

privatenumber:
`source-map-support` was originally added for source map support in older versions of Node.js. Now it's not necessary.

I was thinking it may still be used in cases where users set their own `Error.prepareStackTrace` (e.g. https://github.com/esbuild-kit/esm-loader/blob/944cde52cfb6f6a72d45a8f0159b68c97468924b/tests/specs/typescript/ts.ts#L36), but I'm not sure if this is common. Will test further with `source-map-support` but I don't believe this is a breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,7 @@
 	},
 	"dependencies": {
 		"esbuild": "~0.18.20",
-		"get-tsconfig": "^4.7.2",
-		"source-map-support": "^0.5.21"
+		"get-tsconfig": "^4.7.2"
 	},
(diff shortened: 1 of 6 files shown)


privatenumber:
I ran it with `-r source-map-support/register` which overrides `Error.prepareStackTrace` and seems it passes tests. Should be safe to move forward with.

privatenumber:
:tada: This issue has been resolved in v4.3.0

Issue #257, opened 2023-06-15
Support TypeScript extensions when using `--test` flag.


remcohaszing:
### Feature request

Running `node --test` [discovers test files](https://nodejs.org/api/test.html#running-tests-from-the-command-line). However, it doesn’t discover files that end with the `.ts`, `.cts`, `.mts`, or `.tsx` extensions. It would be nice if `tsx` supports this.

### Why?

I would like to leverage the builtin Node test runner. I also like using `tsx`. They work great together, except for this quirk meaning I need to specify all tests as command line parameters.

### Alternatives

The workaround is to specify test files as command line parameters. This works, but it’s not ideal.

### Additional context

https://nodejs.org/api/test.html#test-runner-execution-model

privatenumber:
Happy to accept a PR to add support.

IIRC there isn't an API to extend this. I recommend starting by opening a feature request with Node.js.

bestickley:
My current workaround is: `glob './src/**/*.test.ts' --all --cmd='tsx --test'` which requires an additional dependency: [glob](https://www.npmjs.com/package/glob)

kasajian:
> My current workaround is: `glob './src/**/*.test.ts' --all --cmd='tsx --test'` which requires an additional dependency: [glob](https://www.npmjs.com/package/glob)

This solves the problem even on Windows.

Add this to your scripts section:

        "test": "glob ./test/**/*.test.?s --all --cmd=\"tsx --test\"",

If you don't already have glob as a dependency, add it as a Dev dependency:

        npm i -D glob

privatenumber:
This issue is for implementing a feature, and these workarounds not productive towards that goal so I'm going to hide them.

For anyone encountering this problem:

We're looking for a working approach that actually extends Node.js' `--test` glob.

Personally, I don't use Node's `--test` feature so I'm not incentivized to prioritize this. But I'm happy to prioritize and support anyone that wants to take initiative on this.

privatenumber:
Investigated this.

Seems like the `node --test` flag accepts globs, so you can simply pass in one that catches TypeScript files.

Since everything passed into tsx is passed down to Node, you can also pass it into tsx:
```sh
tsx --test '**/*.test.ts'
```

(Note: quoting the glob prevents your shell from expanding it so Node will expand it instead)

The default glob Node uses is here:
https://github.com/nodejs/node/blob/v21.2.0/lib/internal/test_runner/utils.js#L51

[...]

PR #410, opened 2023-11-23
feat: support TypeScript in `--test` flag

privatenumber:
fixes https://github.com/privatenumber/tsx/issues/257

This adds TypeScript file detection to Node v21 and higher. Older versions are not supported because it requires a glob file-traverser and it sounds like quite a bit of implementation & maintenance work when users can just upgrade Node.


privatenumber:
:tada: This issue has been resolved in v4.4.0

Issue #218, opened 2023-04-08
Add logging to watch function for easier troubleshooting


Amorites:
### Feature request

It would be helpful to have more detailed logging when using the watch function in TSX. Specifically, it would be beneficial to log the specific event and path that caused a change, making it easier to troubleshoot issues. This would greatly reduce the time and effort required to identify and resolve problems.

### Why?

While working with TSX in conjunction with Vite SSR, I have encountered an issue where Vite generates intermediate files during startup, causing unnecessary reruns that have caused me to spend a significant amount of time troubleshooting. Despite attempts to add logging to the TSX rerun function, I have found that the log function returns "undefined". I eventually discovered that the `debounce` process was discarding all parameters, leading to this issue.

This issue has caused significant disruptions to my workflow, as it has made it difficult to identify and resolve issues efficiently. It would be beneficial to provide a solution for this problem, such as modifying the debounce process or providing a workaround for the logging issue.

mattfysh:
I'm also having trouble getting watch to work and manually added some log lines to dist/cli.js

Specifically I can see a whole bunch of node_modules files in the .on('message') callback, but nothing from my local project dir. As a result, watch will only rerun if I update the entry file, but does not rerun for all other files

I'm not sure why my other files don't come through from `runProcess.on('message')` but I've had to revert to nodemon for the time being: https://github.com/esbuild-kit/tsx/blob/64ece39ea4ecda6b103901b0fbb39799d217dce3/src/watch/index.ts#L86

privatenumber:
If you think there's a bug, would you mind filing a bug report with a repro?

karlhorky:
@privatenumber Nice, thanks for #412! I'll test it with my failure case when it's released (potentially in `4.5.0`)

karlhorky:
Ok, looking at the logs with [`tsx@4.5.0`](https://github.com/privatenumber/tsx/releases/tag/v4.5.0), I see that it's listening to the compiled files in the `.next` folder, which I imagined would have been ignored because [hidden `.*` directories are ignored by default](https://github.com/privatenumber/tsx#:~:text=All%20imported%20files%20are%20watched%20except%20from%20the%20following%20directories%3A%20node_modules%2C%20bower_components%2C%20vendor%2C%20dist%2C%20and%20.*%20(hidden%20directories).) (but maybe they are not ignored when they are part of a path in another folder).

```
[...]

PR #412, opened 2023-11-23
feat(watch): log reason for rerun & improved exit handling

privatenumber:
fix https://github.com/privatenumber/tsx/issues/218

In addition, this improves child process exit handling:
- When a reload is triggered, it sends `SIGTERM` to the child, but if it doesn't exit within 5s, it sends a `SIGKILL`
- When the watcher is exiting (e.g. Ctrl+C), it sends a `SIGKILL` if it doesn't exit within 5s
  - The rationale here is that you're quitting the watcher instead of the process. And it doesn't make sense that the child process can trap the watcher and prevent it from exiting.
- Double Ctrl+C to force quit

### DIFF
--- a/src/esm/register.ts
+++ b/src/esm/register.ts
@@ -9,10 +9,7 @@ export const registerLoader = () => {
 	if (process.send) {
 		port1.addListener('message', (message) => {
 			if (message.type === 'dependency') {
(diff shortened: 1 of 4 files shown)


privatenumber:
:tada: This issue has been resolved in v4.5.0

PR #418, opened 2023-11-28
fix: clarify Node flag claim in README

michaelangeloio:
This doesn't work:

```
npx tsx foo.ts --tsconfig tsconfig.json --env-file=.env     
```

This does work:

```
 npx tsx --env-file=.env  foo.ts  --tsconfig tsconfig.json
```


privatenumber:
FYI this is the same behavior as Node so I was thinking it would be expected behavior

michaelangeloio:
Perhaps it's best to be specific? 🤷 Up to you, just thought it would help the next guy/gal

PR #475, opened 2024-02-10
fix: resolve tsconfig paths inside node_modules

### LINKED ISSUE #159: Resolving tsconfig paths breaks inside node_modules
### Bug description

I put my project inside `node_modules` directory (not even real `node_modules` alongside some `package.json`, just a new directory named that).

After doing so, resolving tsconfig paths stopped working.

I expect tsx to be agnostic to where the project is located (as long as it has its own `package.json` and `tsconfig.json`), same as esbuild is.

### Reproduction

I repeated this on Stackblitz: https://stackblitz.com/edit/node-qjlxvn

This does not completely resemble the description above (it uses a 'real' node_modules directory) but the very same code can be used to repeat the problem on a completely empty node_modules.

Run:

```ts
~/projects/node-qjlxvn
❯ cd test

~/projects/node-qjlxvn/test
❯ tsx index.ts
value

~/projects/node-qjlxvn/test
❯ cd ..

~/projects/node-qjlxvn
❯ mv test node_modules/

~/projects/node-qjlxvn
❯ cd node_modules/test

~/projects/node-qjlxvn/node_modules/test
❯ tsx index.ts
Error: Cannot find module '@/value'
Require stack:
- /home/projects/node-qjlxvn/node_modules/test/index.ts
    at Module._resolveFilename (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:6:217308)
    at d.default._resolveFilename (file:///home/projects/node-qjlxvn/node_modules/.pnpm/@esbuild-kit+cjs-loader@2.4.1/node_modules/@esbuild-kit/cjs-loader/dist/index.js:1:1554)
    at Module._load (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:6:214847)
    at Module.require (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:6:218087)
    at i (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:6:415284)
    at _0xc8b09a (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:15:142803)
    at eval (file:///home/projects/node-qjlxvn/node_modules/test/index.ts:2:774)
    at Object.eval (file:///home/projects/node-qjlxvn/node_modules/test/index.ts:3:3)
    at Object.function (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:15:143540)
    at Module._compile (https://nodeqjlxvn-s25i.w-credentialless.staticblitz.com/blitz.b6c96f782a49b3e017dca41830943768f8acbe40.js:6:219079) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/home/projects/node-qjlxvn/node_modules/test/index.ts' ]
}

~/projects/node-qjlxvn/node_modules/test
❯ esbuild --bundle index.ts
(() => {
  // src/value.ts
  var value_default = "value";

  // index.ts
  console.log(value_default);
})();
```

### Environment

```shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    tsx: ^3.12.1 => 3.12.1
```

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

shigma:
fix #159

I commented out the lines regarding `node_modules` detection. Already tested in my environment.

However I'm not sure what the original intention of this line of code is (performance concern?), or whether it breaks anything else.

### DIFF
--- a/src/esm/loaders.ts
+++ b/src/esm/loaders.ts
@@ -152,7 +152,8 @@ export const resolve: resolve = async function (
 	if (
 		tsconfigPathsMatcher
 		&& !isPath // bare specifier
-		&& !context.parentURL?.includes('/node_modules/')
+		// https://github.com/privatenumber/tsx/issues/159
+		// && !context.parentURL?.includes('/node_modules/')
 	) {
 		const possiblePaths = tsconfigPathsMatcher(specifier);
 		for (const possiblePath of possiblePaths) {


privatenumber:
I believe this would apply the tsconfig paths from the project to dependencies, which I think is incorrect behavior.

For example, if there's a tsconfig path called "fetch", and a dependency happens to be importing a package called "fetch", it would be overwritten to resolve using the projects tsconfig path.

shigma:
The scenario you mentioned is exactly what I need. I can give you a detailed explanation.

I am developing two packages (namely A and B for simplicity) in different repositories (I didn't put them in a monorepo because there is not much connection between the two packages, and the repositories containing A and B are already monorepos). A is an underlying library for B (which means B has a dependency of A), and B is a building tool for A (which means A has a dev dependency of B).

In the repository of A, the package manager will point dependency `B > A` to `A` itself. As you can see, the building process of A involves B, and then involves A (which has not been built). It would be perfect if `tsx` respected the tsconfig paths when resolving `B > A`. Unfortunately, it does not.

If you don't think the above scenario should be handled by default, how about adding an environment variable? That would fix #159 as well.

privatenumber:
tsx is only concerned with following TypeScript behavior. It would help if you can demonstrate how TypeScript handles this.

shigma:
It has nothing to do with typescript (if is does, then `node_modules` should not be involved in the judgment). With `tsc -b` my repository will compile without any errors.

It is about the behavior of the package manager. I don't know what the behavior of pnpm is, at least the behavior of yarn is that as long as a package exists in the local workspace, it will not try to install it from the registry (even if the version in the local workspace does not match with the requested version).

Under this premise, as long as any package downloaded from the registry depends on a package that exists in the local workspace, it should be allowed to use tsconfig paths to resolve, otherwise the package will not function properly.

shigma:
- Current behavior: tsconfig paths should not be respected if `parentURL` is under `node_modules`.
- Desired behavior of #159: tsconfig paths should not be respected if the `parentURL` is under `node_modules` of the *current project* (although this may be difficult to judge).
- Desired behavior of above scenario: tsconfig paths should be respected as long as the *resolved URL* is not under `node_modules`.
- Actual behavior of this PR: always respect tsconfig paths, whether under `node_modules` or not.

privatenumber:
> It has nothing to do with typescript

tsx is a TypeScript runtime. And TypeScript resolves `node_modules` too (this is how the type checker works). Still looking to understand TS behavior here.

Also, please try to keep your messages shorter. Sorry but unless you're a sponsor this is just something I do for fun when I have a few min to spare, and I'm less inclined to read long texts.

shigma:
But the parent module is a `.js` file:

```js
// path/to/a/node_modules/b/index.js
import xxx from 'a' // resolve to package located in path/to/a (node behavior)
```

```js
// path/to/a/index.ts
import xxx from 'a' // resolve to package located in path/to/a (node behavior)
```

The above two file request the same package, and should point to the same file.

privatenumber:
TypeScript can handle and read types from .js files (JSdoc) in `node_modules` as well: https://github.com/microsoft/TypeScript/issues/33136#issuecomment-578699134

Try creating a minimal reproduction with a fake `node_modules` dependency,  and run `tsc --traceResolution --allowJs` to see if the tsconfig paths gets applied to the dependency.

- https://www.typescriptlang.org/tsconfig/#traceResolution

privatenumber:
@shigma Any updates?

shigma:
Sorry for the delay. Minimal reproduction: https://github.com/shigma/tsx-475

privatenumber:
To be clear, the reproduction is to understand how TypeScript works.

Can you remove yarn and tsx from the reproduction? And instead, make the `node_modules/custom-pkg` directory yourself that imports the same alias the parent project does, and run `tsc --traceResolution --allowJs` to see how it resolves it.

shigma:
> Can you remove yarn and tsx from the reproduction?

https://github.com/shigma/tsx-475/tree/pure

> make the node_modules/custom-pkg directory yourself that imports the same alias the parent project does

Sorry I don't understand. I'm not using any aliases.

> `tsc --traceResolution --allowJs`

It produces very long output, you may need to inspect it yourself.

privatenumber:
> It produces very long output, you may need to inspect it yourself.

This is your PR so you'll have to do the minimum work of proving it meets TypeScript behavior.

FYI tsx is a result of spending hours and days analyzing that output.

> Sorry I don't understand. I'm not using any aliases.

I'm talking about tsconfig paths (which are aliases).

shigma:
https://github.com/shigma/tsx-475/blob/pure/tsc-output.txt

I didn't use `--traceResolution` before and am not sure if this is what you need.

> I'm talking about tsconfig paths 

The other package is intended to be publish without having any tsconfig. You can verify the behavior of typescript under this premise.

Also sorry for taking up your time. If you are tired of these comments please close this PR. It might be good for you and me if I fork a copy and modify it myself. Hope you have a nice day.

PR #496, opened 2024-03-15
document `require('tsx/cjs')` usage

mmkal:
Hi - I've been gradually moving over from ts-node, and one scenario I noticed wasn't in the readme was an equivalent to `require('ts-node/register'). Useful when you need to use typescript modules from a commonjs script. For example, for a CLI tool like ESLint that doesn't support typescript configs. You can write an [eslint.config.js](https://eslint.org/docs/latest/use/configure/configuration-files-new#configuration-file) like this:

```js
require('tsx/cjs')

module.exports = [
  require('./src/config1.ts'),
  require('./src/config2.ts'),
]
```

This is much easier than figuring out have to run eslint via tsx or pass in a `--require` CLI argument. You can just run `./node_modules/.bin/eslint` normally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,14 @@ If you only need to add TypeScript & ESM support in a CommonJS context, you can 
 node --require tsx/cjs ./file.ts
 ```
 
+You can also `require` the CJS loader from a commonjs script, enabling you to use typescript modules from that script:
+
+```js
+require('tsx/cjs')
+
+module.exports = require('./src/foo.ts')
+```
+
 ### Hashbang
 
 If you prefer to write scripts that doesn't need to be passed into tsx, you can declare it in the [hashbang](https://bash.cyberciti.biz/guide/Shebang).


privatenumber:
I think that's kind of a hacky solution to your problem because you're sneaking it in via config—exploiting something that's only be available if the CLI accepts a JS config.

But Node actually offers an official API for this that works for any CLI (which is documented), even if it doesn't offer a JS config:

```
NODE_OPTIONS='--require tsx/cjs' eslint .
```

Or

```
NODE_OPTIONS='--import tsx' eslint .
```

mmkal:
I think they're about equal in hackiness. When using a js config (note that eslint only supports js flat config now), it's better to have a config file that just works rather than require a special environment variable that some might forget to use, or forget to remove when it's no longer needed. 

Basically I think it's a pretty valid use case, and it's up to end users whether it's better for them. And tsx already supports it, so worth documenting.

Up to you though!

Max10240:
Lucky to see this, as far as I'm concerned, this method works very well:
I only want to import a ".ts" configuration file into a ".js" file, just like a "ts-config-loader", the only thing I need to do is add this line "require('tsx/cjs')" into the ".js" file. cheer
like:
```tsx
// @filename: foo.config.ts
export default {
  foo: {
  }
} satisfies { foo: unknown }
```

```js
// @filename: bar.js
require('tsx/cjs');

const CONFIG_PATH = '.../foo.config.ts';
const config = require(CONFIG_PATH); // works well
```

Issue #506, opened 2024-03-27
VS Code breakpoint with tsx and Nodejs 20.12.0 does not work


piquark6046:
### Acknowledgements

- [X] I searched existing issues to avoid duplicates
- [X] I understand this is a place to report a confirmed bug (not seek debugging help)
- [X] I understand this is a collaborative open source project, and relies on community contributions
- [X] I have read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/develop/CONTRIBUTING.md)

### Minimal reproduction URL

https://github.com/List-KR/multithread-array/pull/1

### Version

4.7.1

### Node.js version

20.12.0

### Package manager

npm

### Operating system

Linux

### Problem & expected behavior (under 200 words)

- What happened: All VS Code breakpoints are ignored
- What I expected: VS Code breakpoint works as using Node.js 20.11.1
- Screenshot:
  <details><summary>Screenshot</summary>
  <p>
  
  ![image](https://github.com/privatenumber/tsx/assets/98787049/8ee25c54-f699-4b82-a263-8782b1d63847)
  
  
  </p>
  </details> 

- devcontainer.json
  ```json
  {
	  "name": "Typescript with Rust",
	  "image": "mcr.microsoft.com/devcontainers/base:ubuntu",
	  "features": {
		  "ghcr.io/devcontainers/features/node:latest": {
			  "nodeGypDependencies": true,
			  "version": "lts"
		  },
		  "ghcr.io/devcontainers/features/rust:latest": {
			  "version": "latest",
			  "profile": "default"
		  },
		  "ghcr.io/devcontainers-contrib/features/bash-command:latest": {
			  "command": "echo hi!"
		  },
		  "ghcr.io/devcontainers-contrib/features/npm-package:latest": {
			  "package": "typescript",
			  "version": "latest"
		  },
		  "ghcr.io/devcontainers-contrib/features/typescript:latest": {
			  "version": "latest"
		  },
		  "devwasm.azurecr.io/dev-wasm/dev-wasm-feature/rust-wasi:latest": {}
	  },
	  "remoteUser": "root",
	  "mounts": [
		  "source=${localEnv:HOME}/repos/Filters,target=/workspaces/resources/Filters,type=bind,consistency=cached",
		  "source=${localEnv:HOME}/repos/Userscript,target=/workspaces/resources/Userscript,type=bind,consistency=cached"
	  ]
  }
  ```

### Contributions

- [ ] I will open a pull request for this issue
[...]

PR #517, opened 2024-04-03
fix: VS Code breakpoint with tsx and Nodejs 20.12.0 does not work

piquark6046:
Fix https://github.com/privatenumber/tsx/issues/506

### DIFF
--- a/src/source-map.ts
+++ b/src/source-map.ts
@@ -1,6 +1,8 @@
+import * as Semver from 'semver';
 import type { Transformed } from './utils/transform/apply-transformers.js';
 
 const inlineSourceMapPrefix = '\n//# sourceMappingURL=data:application/json;base64,';
+const isPrepareStackTraceExposed = Semver.satisfies(process.versions.node, '>=20.12.0');
 
 export const installSourceMapSupport = () => {
 	const hasNativeSourceMapSupport = (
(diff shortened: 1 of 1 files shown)


src/source-map.ts
```diff
@@ -1,6 +1,8 @@
+import * as Semver from 'semver';
 import type { Transformed } from './utils/transform/apply-transformers.js';
 
 const inlineSourceMapPrefix = '\n//# sourceMappingURL=data:application/json;base64,';
+const isPrepareStackTraceExposed = Semver.satisfies(process.versions.node, '>=20.12.0');
```
privatenumber, line 5: Versioned features live here: https://github.com/privatenumber/tsx/blob/develop/src/utils/node-features.ts

Instead of `semver`, use the `isFeatureSupported` function

src/source-map.ts
```diff
@@ -18,7 +20,7 @@ export const installSourceMapSupport = () => {
 		 *
 		 * https://github.com/nodejs/node/blob/91193825551f9301b6ab52d96211b38889149892/lib/internal/errors.js#L141
 		 */
-		&& typeof Error.prepareStackTrace !== 'function'
+		&& isPrepareStackTraceExposed ? typeof Error.prepareStackTrace === 'function' : typeof Error.prepareStackTrace !== 'function'
```
privatenumber, line 23: Confused what's happening here. It's checking if it's a function, and then if it's not a function.

What are we actually trying to check for here? Can you update the comment?

privatenumber (changes requested):
Thanks for the PR!

Can you pull in https://github.com/privatenumber/tsx/pull/513 for tests?

piquark6046:
I close my PR and please continue in https://github.com/privatenumber/tsx/pull/518.

Issue #433, opened 2023-12-08
Coverage with c8 and tsx reporting as 100% since version 4.3.0


OscarBarrett:
### Acknowledgements

- [X] I searched existing issues before opening this one to avoid duplicates
- [X] I understand this is not a place for seek help, but to report a bug
- [X] I understand that the bug must be proven first with a minimal reproduction
- [X] I will be polite, respectful, and considerate of people's time and effort

### Minimal reproduction URL

https://github.com/OscarBarrett/tsx-coverage-issue

### Version

v4.6.2

### Node.js version

v20.10.0

### Package manager

yarn

### Operating system

Linux

### Problem & Expected behavior

We use c8 and tsx together when running our test suites, and since tsx version 4.3.0 any file that is imported in our tests gets 100% coverage.
Locking to tsx 4.2.1, we get the expected coverage.

Assuming this is related to #405

### Contributions

- [ ] I plan to open a pull request for this issue
- [ ] I plan to make a financial contribution to this project

andreashuber69:
I have the same issue. FWIW, you can work around this by running tests on the built js output. If you have TypeScript generate .map files, then the CC output should look the same as before (i.e. reported on ts files rather than js files). Here's an example of the changes necessary (besides building with tsc):

https://github.com/andreashuber69/verify-coldcard-dice-seed/commit/35b72b9bcc4a087790848db0afb3df9407b0d919

jbergstroem:
Ran into the same; just wanted to chip in and confirm that downgrading works.

privatenumber:
I haven't investigated this but I'm thinking this is as simple as removing this check so sourcemaps are always inlined:
https://github.com/privatenumber/tsx/blob/8c778919496a8164ec63d0b488c805bce1aba839/src/source-map.ts#L23

There will need to be tests though. PR welcome if anyone wants to tackle.

kallekj:
Test coverage works in 4.7.0 if I run mocha with source-map-support/register. 
My `.mocharc` file to run it with tsx/esm:
```json
{
  "import": "tsx/esm",
  "require": "source-map-support/register"
}

```

cenfun:
This issue is related to the source maps conversion with [`v8-to-istanbul`](https://github.com/istanbuljs/v8-to-istanbul), and it is not related to tsx. see [here](https://github.com/jestjs/jest/issues/11188)
V8 provides the correct coverage data
```js
[
[...]

Issue #340, opened 2023-08-23
External source map is not supported


BenceSzalai:
### Problem

Dependency files that:
- include ESM dynamic imports
- has sourcemaps associated (generated by TSC)

show up in Error stack traces:
- with `.js` extension (as opposed to `.ts`)
- with wrong line numbers
- and with their real file path instead of the original path based on the sourcemap

i.e. their source maps are ignored.

### Expected behavior

All dependency files that has sourcemaps associated should show up in Error stack traces with their original path, extension and line numbers.

### Minimal reproduction URL

https://github.com/BenceSzalai/tsx-sm-issue

### Version

v2.5.5

### Node.js version

v18.17.0 / v20.5.1 (v16 is not affected)

### Package manager

npm

### Operating system

macOS

### Contributions

- [ ] I plan to open a pull request for this issue
- [ ] I plan to make a financial contribution to this project

### Notes

Originally [opened against TXS](https://github.com/esbuild-kit/tsx/issues/289), but it looks like it is an issue with ESM Loader.

SystemParadox:
Due to the fact that tsx doesn't support project references yet (https://github.com/privatenumber/tsx/issues/96) we tried to add our own loader on top of tsx that would esbuild each of our monorepo packages on demand. But then we run into this problem and none of the monorepo packages get source maps (confirmed that it does work ok in node 16).

privatenumber:
I tried testing the latest tsx with this but the reproduction was actually not minimal.

Can you provide a StackBlitz reproduction with just one file with an inline source map, demonstrating it doesn't work with tsx?

privatenumber:
:tada: This issue has been resolved in v4.11.1

If you appreciate this project, please consider [supporting this project by sponsoring](https://github.com/sponsors/privatenumber) :heart: :pray:

PR #518, opened 2024-04-04
fix issue 506

cenfun:
fix https://github.com/privatenumber/tsx/issues/506

### DIFF
--- a/src/source-map.ts
+++ b/src/source-map.ts
@@ -1,4 +1,5 @@
 import type { Transformed } from './utils/transform/apply-transformers.js';
+import { isFeatureSupported, prepareStackTraceWithSourcemap } from './utils/node-features.js';
 
 const inlineSourceMapPrefix = '\n//# sourceMappingURL=data:application/json;base64,';
 
@@ -18,7 +19,7 @@ export const installSourceMapSupport = () => {
 		 *
 		 * https://github.com/nodejs/node/blob/91193825551f9301b6ab52d96211b38889149892/lib/internal/errors.js#L141
 		 */
-		&& typeof Error.prepareStackTrace !== 'function'
+		&& (isFeatureSupported(prepareStackTraceWithSourcemap) || typeof Error.prepareStackTrace !== 'function')
 	);
 
 	if (hasNativeSourceMapSupport) {
--- a/src/utils/node-features.ts
+++ b/src/utils/node-features.ts
@@ -53,3 +53,10 @@ export const importAttributes: Version[] = [
 export const testRunnerGlob: Version[] = [
 	[21, 0, 0],
 ];
+
+// https://nodejs.org/docs/latest-v20.x/api/cli.html#--enable-source-maps
+// https://github.com/nodejs/node/pull/50827
+// Allow using --enable-source-maps together with custom Error.prepareStackTrace
+export const prepareStackTraceWithSourcemap: Version[] = [
+	[20, 12, 0],
+];


privatenumber:
Can you pull in this in to see if it passes? https://github.com/privatenumber/tsx/pull/513

cenfun:
I don't know hot to do it, but create PR here https://github.com/privatenumber/tsx/pull/519

src/utils/node-features.ts
```diff
@@ -53,3 +53,10 @@ export const importAttributes: Version[] = [
 export const testRunnerGlob: Version[] = [
 	[21, 0, 0],
 ];
+
+// https://nodejs.org/docs/latest-v20.x/api/cli.html#--enable-source-maps
+// https://github.com/nodejs/node/pull/50827
+// Allow using --enable-source-maps together with custom Error.prepareStackTrace
+export const prepareStackTraceWithSourcemap: Version[] = [
+	[20, 12, 0],
```
privatenumber, line 61: Also `21.6.0`

Seems https://github.com/nodejs/node/pull/50827 is included in the following releases (ordered by release date):
`v21.6.0` https://github.com/nodejs/node/commit/291265ce27
`v21.6.1` https://github.com/nodejs/node/commit/291265ce27
`v21.6.2` https://github.com/nodejs/node/commit/291265ce27
`v21.7.0` https://github.com/nodejs/node/commit/291265ce27
`v21.7.1` https://github.com/nodejs/node/commit/291265ce27
`v20.12.0` https://github.com/nodejs/node/commit/f4d7f0498e
`v21.7.2` https://github.com/nodejs/node/commit/291265ce27
`v20.12.1` https://github.com/nodejs/node/commit/f4d7f0498e

src/source-map.ts
```diff
@@ -18,7 +19,7 @@ export const installSourceMapSupport = () => {
 		 *
 		 * https://github.com/nodejs/node/blob/91193825551f9301b6ab52d96211b38889149892/lib/internal/errors.js#L141
 		 */
-		&& typeof Error.prepareStackTrace !== 'function'
+		&& (isFeatureSupported(prepareStackTraceWithSourcemap) || typeof Error.prepareStackTrace !== 'function')
```
privatenumber: Upon reading what [the default `prepareStackTrace`](https://github.com/nodejs/node/blob/dd711d221a6cd3ffaccedf6fc810b9c2d16eedd5/lib/internal/errors.js#L95-L111) looks like, I'm okay with removing this check entirely.

The Node docs previously implied that if it was overridden sourcemaps wouldn't work expectedly, but the implementation is so simple that the default behavior can easily be replicated.

cenfun: done

src/source-map.ts
```diff
@@ -18,7 +19,7 @@ export const installSourceMapSupport = () => {
 		 *
 		 * https://github.com/nodejs/node/blob/91193825551f9301b6ab52d96211b38889149892/lib/internal/errors.js#L141
 		 */
-		&& typeof Error.prepareStackTrace !== 'function'
+		&& (isFeatureSupported(prepareStackTraceExposed) || typeof Error.prepareStackTrace !== 'function')
```
privatenumber, line 22: Upon reading what [the default `prepareStackTrace`](https://github.com/nodejs/node/blob/dd711d221a6cd3ffaccedf6fc810b9c2d16eedd5/lib/internal/errors.js#L95-L111) looks like, I'm okay with removing this check entirely.

The Node docs previously implied that if it was overridden sourcemaps wouldn't work expectedly, but the implementation is so simple that the default behavior can easily be replicated.

cenfun: are you sure remove checking Error.prepareStackTrace? I don't know the history.

privatenumber:
I'm curious if this would also fix:
- https://github.com/privatenumber/tsx/issues/433
- https://github.com/privatenumber/tsx/issues/340

Will investigate in a bit

cenfun:
for #433 see my comments https://github.com/privatenumber/tsx/issues/433#issuecomment-2011778314

src/source-map.ts
privatenumber: Hmm, I dug into it deeper...

Previously, the Node docs said:
> Overriding `Error.prepareStackTrace` prevents `--enable-source-maps` from modifying the stack trace.

And the `prepareStackTrace` logic was:
- no source maps: https://github.com/nodejs/node/blob/v20.11.1/lib/internal/errors.js#L119-L133
- source maps: https://github.com/nodejs/node/blob/v20.11.1/lib/internal/source_map/prepare_stack_trace.js#L32-L110

Not generating source maps when `Error.prepareStackTrace` was set was a performance optimization because I assumed the custom `prepareStackTrace` wouldn't be able to handle the source map. But this may have been an incorrect assumption. Furthermore, it's a rare optimization because I think `Error.prepareStackTrace` is not very common.

That said, it's probably fine to remove this check entirely for how little benefit it provides, and the possibility that a custom `prepareStackTrace` can handle source maps.

cenfun: removed

privatenumber:
:tada: This issue has been resolved in v4.7.2

SuperchupuDev:
this is still not working for me in some places using the node 21.7.2 test runner, from what i can see it's fixed everywhere except here (using `node --import tsx --test test/*.test.ts`)
image

(it works everywhere else)
image

PR #522, opened 2024-04-12
docs: Add example for hashbang with local tsx

silverwind:
There is also a [variant](https://stackoverflow.com/a/52979955/808699) with `-S`:

```bash
#!/usr/bin/env -S npx tsx
```

but as per my testing using BSD and GNU `env`, the two variants work exactly the same, so I guess without the `-S` would be slightly more portable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,15 @@ $ ./file.ts hello
 argv: [ 'hello' ]
 ```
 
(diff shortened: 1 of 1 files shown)


privatenumber:
Thanks! I appreciate it

privatenumber:

PR #535, opened 2024-04-24
fix: support TS resolution in JS files when `allowJs` is set

### LINKED ISSUE #135: TypeScript resolution not working from JS files (even with `allowJs`)
### Bug description

`tsx index.mts` to run

`index.mts` imports `file.mjs` by `import x from "./file.mjs"`

`file.mjs` imports `rootfile.mts` by `import x from "./rootfile.mjs"`

I expect this to resolve `rootfile.mts` and work.

```
❯ tsx index.mts
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/projects/rootfile.mjs' imported from /home/projects/node-war8wy/file.mjs
    at get (https://node-war8wy.w-corp.staticblitz.com/blitz.a0ea75dd36396805bd36e49d0b54d86602583e15.js:6:292488)
    at U@file:///home/projects/node-war8wy/node_modules/ (esbuild-kit/esm-loader/dist/index.js:19:1634)
    at <anonymous> (<anonymous>) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

### Reproduction

[https://stackblitz.com/edit/node-war8wy](https://stackblitz.com/edit/node-war8wy) has the reproduction

### Environment

```
  System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 7.17.0 - /usr/local/bin/npm
  npmPackages:
    tsx: ^3.11.0 => 3.11.0 
```

### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

nwalters512:
Closes #135.

I tested this locally and it works as expected in my project. I also added a new test which fails on `develop` but passes after these changes.


tests/specs/smoke.ts
nwalters512, line 1: I didn't fully understand how the types are structured/split up; I'd be more than happy to adjust things here in response to feedback.

privatenumber: For context, the idea behind the smoke tests is to test with minimal number of child processes; the other tests are slow primarily because it spawns so many children.

tests/specs/smoke.ts
```diff
@@ -218,6 +218,14 @@ const files = {
 
 	'file.txt': 'hello',
 
+	'import-typescript-parent.js': sourcemap.tag`
+	import './import-typescript-child.js';
```
privatenumber, line 222: Can you move this inside `import-from-js.js`?

tests/specs/smoke.ts
```diff
@@ -218,6 +218,14 @@ const files = {
...
+	`,
+
+	'import-typescript-child.ts': sourcemap.tag`
+	console.log('imported');
```
privatenumber, line 226: Make this `allowJs` or something more unique that can be matched from within the `'from .js'` test

privatenumber:
Really appreciate this PR, thanks @nwalters512 !

BTW I had a vision to leverage get-tsconfig like this https://github.com/esbuild-kit/cjs-loader/pull/34/files but I think there are other problems with that approach and this is a good solution for now. 🙌

tests/specs/smoke.ts
nwalters512: I can, though that will require adding `--tsconfig tsconfig/tsconfig-allowJs.json` to that test's arguments, at which point there wouldn't be any tests that cover the case where TS resolution rules _aren't_ followed. If that's good with you, I'll do the move and delete the `allowJs in tsconfig.json` test.

privatenumber: Ah good point, let's keep it this way then. Thanks for catching it!

privatenumber:

Issue #543, opened 2024-05-05
Modules imported dynamically from an absolute path are not loaded/handled by tsx v4.9.1


fardjad:
### Acknowledgements

- [X] I read the documentation and searched existing issues to avoid duplicates
- [X] I understand this is a place to report a confirmed bug (not seek debugging help)
- [X] I understand this is a collaborative open source project, and relies on community contributions
- [X] I have read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/develop/CONTRIBUTING.md)

### Minimal reproduction URL

https://stackblitz.com/edit/tsx-4-9-1-import-from-absolute-path-issue

### Version

4.9.1

### Node.js version

v20.12.2

### Package manager

npm

### Operating system

macOS

### Problem & expected behavior (under 200 words)

- What happened

The newly released version v4.9.1 doesn't handle absolute paths (4.9.0 doesn't have this issue).

- What I expected

I expected to be able to run the following with `tsx index.mts`:

```typescript
// index.mts

import path from "node:path";
import { fileURLToPath } from "node:url";
import fs from "node:fs";

const __dirname = path.dirname(fileURLToPath(new URL(import.meta.url)));
const moduleRelativePath = "./hello.mts";
const moduleAbsolutePath = path.resolve(__dirname, moduleRelativePath);

fs.writeFileSync(moduleAbsolutePath, `export const message = "Hello, world!";`);

await import(moduleAbsolutePath); // Doesn't work
```

- Error stack trace

```
node:internal/process/esm_loader:34
      internalBinding('errors').triggerUncaughtException(
                                ^
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".mts" for <redacted>/hello.mts
    at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:160:9)
    at defaultGetFormat (node:internal/modules/esm/get_format:203:36)
    at defaultLoad (node:internal/modules/esm/load:143:22)
    at async nextLoad (node:internal/modules/esm/hooks:866:22)
    at async x (file:///<redacted>/.npm/_npx/dcefadc2074ecd1f/node_modules/tsx/dist/esm/index.mjs?1714902597374:2:1744)
    at async nextLoad (node:internal/modules/esm/hooks:866:22)
    at async Hooks.load (node:internal/modules/esm/hooks:449:20)
    at async handleMessage (node:internal/modules/esm/worker:196:18) {
[...]

PR #544, opened 2024-05-05
fix(esm): resolve absolute paths

fardjad:
This PR attempts to fix the issue described here: https://github.com/privatenumber/tsx/issues/543

### DIFF
--- a/src/esm/hook/resolve.ts
+++ b/src/esm/hook/resolve.ts
@@ -121,6 +121,35 @@ export const resolve: resolve = async (
 		return nextResolve(specifier, context);
 	}
 
+	const isPath = (
+		specifier.startsWith(fileProtocol)
+		|| isRelativePathPattern.test(specifier)
+	);
+
+	// bare specifier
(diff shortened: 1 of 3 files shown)


privatenumber:
@fardjad Would you mind adding a test?

fardjad:
@privatenumber I wasn't sure where would be the best place to add the test. I updated my PR. Do you think the approach and the place for adding the test make sense or do you have something else in mind?

fardjad:
@privatenumber Of course! I updated my PR.

I wasn't sure where the best place to add the test would be. I decided to put it under `register / unregister` group as it affects the resolver (and not `tsImport` itself). Do you think the approach and the location for adding the test make sense, or do you have something else in mind?

privatenumber:
That works, but I consolidated it into the smoke test—previously, our tests would take super long from spawning so many processes.

Thanks for your contribution!

privatenumber:
:tada: This issue has been resolved in v4.9.2

PR #551, opened 2024-05-13
fix: expose types for `esm/api`

antfu:
I was using `ReturnType<typeof register>` to get the type - which I wasn't able to get the overloads with namespaced one. I think it would be nice to expose them.

### DIFF
--- a/src/esm/api/index.ts
+++ b/src/esm/api/index.ts
@@ -1,2 +1,10 @@
-export { register } from './register.js';
+export {
+	register,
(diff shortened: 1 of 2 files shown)


privatenumber:
Sounds good. I think you'll want to re-export them here too:
https://github.com/privatenumber/tsx/blob/master/src/esm/api/index.ts

privatenumber:
:tada: This issue has been resolved in v4.10.2

PR #573, opened 2024-06-06
docs: remove duplicate `Tracking loaded files` section

aryaemami59:
## **This PR:**

- [X] Removes duplicate [`Tracking loaded files`](https://tsx.is/node/esm#tracking-loaded-files) section from docs.


privatenumber:
Thanks!

PR #592, opened 2024-06-17
fix: minimum Node version in warning for `module.register()`

mnaoumov:
According to https://nodejs.org/api/module.html#moduleregisterspecifier-parenturl-options
Module.register was Added in: v20.6.0, v18.19.0

### DIFF
--- a/src/esm/api/register.ts
+++ b/src/esm/api/register.ts
@@ -37,7 +37,7 @@ export const register: Register = (
 	options,
 ) => {
 	if (!module.register) {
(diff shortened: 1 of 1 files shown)


privatenumber:
Nice catch!

I have it here but looks like I typo'd it when I coped it:
https://github.com/privatenumber/tsx/blob/v4.15.5/src/utils/node-features.ts#L41

privatenumber:
:tada: This issue has been resolved in v4.15.6

If you rely on this project, please consider [making a donation](https://github.com/sponsors/privatenumber) :heart: :pray:

PR #618, opened 2024-07-23
Prefer conditional exports

### LINKED ISSUE #617: resolver prioritizes extension guessing over real path
### Acknowledgements

- [X] I read the documentation and searched existing issues to avoid duplicates
- [X] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [X] I understand this is a free project and relies on community contributions
- [X] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

https://github.com/nwalters512/tsx-dynamic-import-default-repro

I would expect the exact same output when compiling the `src/index.ts` program to `*.js` with `tsc` and running the compiled file with Node, and when running `src/index.ts` directly with `tsx`. Instead, the behavior is different:

- When compiling to JS and executing with Node, the `plugin` object has a `default` property.
- When executing directly with `tsx`, the `plugin` object does not have a `default` property.

I'd be happy to further investigate and possibly fix this if you could point me in the direction of whatever part of `tsx` handles things like this (dynamic imports? ESM compat? I'm not sure exactly what `tsx` does that would be causing this, I had assumed that this would all be strictly Node's responsibility.

- [X] I'm interested in working on this issue

nwalters512:
Closes #617.

### DIFF
--- a/src/utils/map-ts-extensions.ts
+++ b/src/utils/map-ts-extensions.ts
@@ -1,6 +1,10 @@
 import path from 'node:path';
 
-const noExtension = ['.js', '.json', '.ts', '.tsx', '.jsx'];
+// When the file extension is missing in the requested path, always try the
+// file path verbatim before trying to resolve it with specific extensions.
+// This ensures that we give Node a chance to take `"exports"` in `package.json`
+// into account.
+const noExtension = ['', '.js', '.json', '.ts', '.tsx', '.jsx'];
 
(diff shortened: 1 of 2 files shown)


privatenumber:
Thanks for the PR but the implementation turned out a lot more complicated when considering more cases.

Closing in favor of https://github.com/pvtnbr/tsx/pull/67

PR #620, opened 2024-07-23
chore: remove unused `eslintConfig`

nwalters512:
Putting the ESLint config in `eslint.config.js` makes it discoverable by editor integrations so that they can show lint errors/fixes inline.

AFAICT `lintroll` completely ignores the `eslintConfig` property in `package.json`, so it already wasn't necessary (and there wasn't a `@pvtnbr/*` package installed to begin with). The ignore of `tests/fixtures`, despite being ignored in the first place, also just isn't necessary.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,0 +1,11 @@
+// @ts-check
+import { defineConfig, pvtnbr } from 'lintroll';
+
+export default defineConfig([
+	{
+		ignores: ['tests/fixtures'],
+	},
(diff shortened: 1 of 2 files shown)


privatenumber:
Good catch, this should've been removed. The config is just passed in as CLI flags to `lintroll` now.

privatenumber:
This issue is now resolved in [v4.16.3](https://github.com/privatenumber/tsx/releases/tag/v4.16.3).

PR #561, opened 2024-05-25
feat(watch): `--include` & `--exclude` CLI flags

boxizen:
closes #181 

This change adds additional files to the watchlist via adding the `add` parameter.

PR #625, opened 2024-07-31
feat(watch): `--include` CLI flag to watch additional files

### LINKED ISSUE #181: Add files to be watched
### Feature request

I would like to be able to add files to be watched in addition to the default, for example to watch the `.env` file.

### Why?

I use a `.env` file with `dotenv` for environment variables. Often I'll change a value in `.env` during runtime to test something (e.g. enable/disable feature flag). 

With nodemon I'm using:

```
nodemon --watch src --watch .env src/index.ts
```

It would be nice to be able to do this with tsx as well.

### Alternatives

My current alternative is nodemon+ts-node+swc. But tsx seems like a more convenient all-in-one solution.

### Additional context

_No response_

mdmower-csnw:
_This PR picks up from where #561 left off (it appears to have been abandoned), but there is actually very little of that PR remaining here. Notably, I didn't see it necessary to add `--exclude` which would only alias the `--ignore` flag. Nonetheless, the chain of commits here reflects the original contribution._

- Add support for watch CLI flag `--include` that can be used to include additional file paths and glob patterns (same syntax as the `--ignore` flag).
- Add test that ensures both file path and glob pattern work with `--include` flag.

Fixes #181


tests/specs/watch.ts
```diff
@@ -300,5 +303,57 @@ export default testSuite(async ({ describe }, { tsx }: NodeApis) => {
...
+						await fixture.writeFile(entryFile, 'console.log("TERMINATE")');
+					} else if (chunkString.includes('TERMINATE')) {
+						tsxProcess.kill();
+					}
+				});
+
+				const tsxProcessResolved = await tsxProcess;
+				const stdout = stripAnsi(tsxProcessResolved.stdout);
+				expect(stdout).toContain(`change in ./${fileA}`);
+				expect(stdout).toContain(`change in ./${fileB}`);
```
mdmower-csnw, line 353: Need to handle mixed path separators on windows. From test output:
```
Expected substring: "change in ./directory/file-b"
Received string:    "content-a content-b
12:18:14 AM [tsx] change in ./file-a Rerunning...
12:18:15 AM [tsx] change in ./directory\\file-b Restarting...
update-a update-b
12:18:16 AM [tsx] change in ./index.js Rerunning...
TERMINATE"
```

Would also be good to increase timeout between writing fileA and fileB since I'm not seeing `update-a content-b` in the output.

mdmower-csnw:
Thanks for all the CI reruns @privatenumber (and also sorry you had to keep rerunning). I had some trouble installing node-gyp on Windows.

mdmower-csnw:
@privatenumber - is there anything else you need from me for this PR?

privatenumber:
Sorry, hard to prioritize tsx these days. But eventually I'll have to figure out funding first so it makes sense for me to invest more time on tsx, and then I plan to work on a bunch of watch improvements including this.

privatenumber:
I was debating between naming it `watch` vs `include`, but I think `include` makes sense to avoid the double "watch".
I think I'll have to rename `ignore` to `exclude` in the next major.

privatenumber:
This issue is now resolved in [v4.18.0](https://github.com/privatenumber/tsx/releases/tag/v4.18.0).

PR #643, opened 2024-08-31
fix: Drop default excludes that are exist in includes

lbennett-stacki:
Preserves the intention of users that explicitly specify an included path that may exist in the default excludes list.

For example, the following command should trigger a re-run for any changes in node_modules/linked-dependency, despite node_modules appearing in the default excludes.

```bash
tsx watch --include "node_modules/linked-dependency/**/*" src/index.ts
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,6 +77,7 @@
 		"node": ">=18.0.0"
 	},
 	"dependencies": {
+		"anymatch": "^3.1.3",
 		"esbuild": "~0.23.0",
 		"get-tsconfig": "^4.7.5"
 	},
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,9 @@ importers:
 
   .:
     dependencies:
+      anymatch:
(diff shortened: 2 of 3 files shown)


lbennett-stacki:
I recognise there is probably a better method here, around either:

1. Resolve all default excluded path patterns, then only remove the paths that match the include pattern.

```ts
defaultExcludes = resolveAll(defaultExcludes).filter(exclude => doesNotMatchAnyInclude(exclude));
```

2. Manipulating exclude strings to negate any includes from the excludes list

```ts
const includes = ["node_modules/linked-dependency/**/*"];
// results in...
excludes[0] === "**/node_modules/**"
// manipulated into...
excludes[0] === "**/node_modules/!(linked-dependency)/**"

```

However I think this PR is a simple first step whilst a decision is made by maintainers. If users are setting --include flags, their intention should be preserved, they are also much more likely at this point to understand the include/exclude functionality so can exclude any paths that end up causing issues themselves.

PR #656, opened 2024-09-24
fix: Allow ignore patterns to apply to absolute path instead of relative path

kingston:
+ Upgrade chokidar to v4 for a smaller footprint
+ Add picomatch for glob matching
+ Add test for parent directory ignores


kingston:
@privatenumber just kicking this off as an initial draft PR so we can comment on raw code vs. ideas. Right now all tests pass except one (--include with glob patterns). This is because chokidar v4 removes all glob support including include patterns. This is a bit trickier to resolve since we'd need to find all the files that match that glob which gets a lot messier/more complex. We can either stay with chokidar v3 or do something a bit more complex. Looking at chokidar v3, it seems like when they encounter a glob pattern they just take the glob-parent and watch the highest parent of the glob e.g. for /blah/foo/**/test - they watch /blah/foo and filter by **/test. So we can try to mimic this behavior manually.

We could potentially do two separate watchers:
1) Executed file + dependencies watcher (basically what we have today with this PR)
2) If --include is present, includes-only watcher. Watch all glob-parents and apply --include globs to filter out only the files we care about. (would coincidentally also address https://github.com/privatenumber/tsx/pull/643)

I don't think it adds too much of an overhead and simplifies the logic.

On a side note, I realized the docs were out of date with --exclude / --include (https://github.com/privatenumber/tsx/blob/master/docs/watch-mode.md) but can be a separate issue to resolve.

privatenumber:
That sounds good to me! Thanks @kingston

kingston:
@privatenumber I've updated the PR to now pass all tests including the include tests. I introduced 2 additional libraries to get the glob parent and check if it is a glob or not (otherwise glob parent would remove the ending).

Overall, I think it should be good to go. However, there's one scenario where I'm unclear what the behavior should be.

If you pass a directory to --exclude without any glob patterns, should it automatically match all its children? e.g. --exclude ./foo will match ./foo/**? If so, we'd probably need to re-work the logic to create a separate matcher for the exact path and path with /** appended to it for non-glob patterns.

Also left a small update the docs to make it more accurate.

kingston:
@privatenumber I don't have access to a windows instance but just uploaded a fix that should address the failing test.

kingston:
Hmm.. that didn't seem to work. I'll try to see if I can get a windows machine to test it on since I have a Mac.

privatenumber:
I actually don't have a Windows machine either. It's not the most ergonomic environment but I just make a separate branch, update the GitHub Actions CI to only run on Windows, and debug there 😅

I'll have more time this weekend so I can take a closer look/help out.

kingston:
@privatenumber OK, I spun up a Windows VM on my Mac and was able to debug more ergonomically :). It looks like globs and Windows paths are complicated since globs are technically not allowed to have \ in them so I used a library normalize-paths (used by chokidar v3 as well) to normalize the paths to / since chokidar v4 also does that when passing to ignored. This makes all tests pass on my Windows 11 VM :).

Fun side-note: I was unable to commit to the repo because of the lint pre-commit hook since the lint command doesn't work on Windows so I had to skip the hooks when committing.

```
C:\...\tsx\docs\getting-started.md
  62:14  error  Parsing error: Unexpected token :

C:\...\tsx\docs\typescript.md
  108:14  error  Parsing error: Unexpected token :
  123:14  error  Parsing error: Unexpected token :
```
  
  My guess is that the glob pattern for ignoring files also doesn't work on Windows :P
  
`  lintroll --node --cache --ignore-pattern 'docs/*.md' .`

kingston:
@privatenumber just wanted to follow up - are there any other action items for this PR?

privatenumber:
Sorry @kingston been swamped lately. I'll try to take a look Sunday or next week.

kingston:
Sure no problem - I get it.

src/watch/index.ts
```diff
@@ -208,29 +234,39 @@ export const watchCommand = command({
...
+			// ignore all files not in includes or explicitly excluded
+			// we need to make sure not to ignore directories otherwise chokidar won't check for it
+			ignored: file => (
+				!globParents.includes(file)
+				&& (
+					!isOptionsInclude(file)
+					|| isOptionsExclude(file)
+				)
+			),
+		}).on('all', reRun);
```
privatenumber, line 267: Is it possible to add to the same watcher above instead of making a new one?

src/watch/utils.ts
```diff
@@ -29,3 +31,10 @@ export const debounce = <T extends (this: unknown, ...args: any[]) => void>(
 		);
 	} as T;
 };
+
+export const resolveGlobPattern = (pattern: string): string => {
+	if (path.isAbsolute(pattern)) {
+		return normalizePath(pattern);
+	}
+	return normalizePath(path.join(process.cwd(), pattern));
+};
```
privatenumber, line 40: This only seems used to normalize the path slashes in user provided paths—do they need _need_ to be normalized?

src/watch/index.ts
```diff
@@ -208,29 +234,39 @@ export const watchCommand = command({
...
-			],
 			ignorePermissionErrors: true,
+			// ignore all files that are by default ignored or explicitly excluded
+			ignored: file => isDefaultIgnore(file) || isOptionsExclude(file),
 		},
 	).on('all', reRun);
 
+	if (resolvedIncludes.length > 0) {
+		const globParents = resolvedIncludes.map(pattern => (
+			isGlob(pattern)
```
privatenumber, line 249: This seems like a pretty [hefty function](https://github.com/micromatch/is-glob/blob/master/index.js) to run only to check if it's a glob. What if we just run `globParent` on all inputs regardless of whether they're a glob?

src/watch/index.ts
```diff
@@ -4,7 +4,10 @@ import { constants as osConstants } from 'node:os';
 import path from 'node:path';
 import { command } from 'cleye';
 import { watch } from 'chokidar';
+import picomatch from 'picomatch';
```
privatenumber, line 7: Picomatch is actually the largest dependency now:
https://pkg-size.dev/chokidar%20glob-parent%20is-glob%20normalize-path%20picomatch

Is there a lighter alternative?

privatenumber:
I pulled out your docs changes into `master` so we can get it out faster. Thanks!

src/watch/index.ts
kingston: I tried that but a simple include like test/foo/hello becomes test/foo when run through glob-parent. We could proactively append a separator at the end but that may result in weirder behavior potentially. Coincidentally glob-parent uses isGlob under the hood.

src/watch/index.ts
kingston: I created a separate one since the logic of the ignore is different for explicit includes compared to default includes (we want to include things that might be by default excluded e.g. node_modules) and I think under the hood the chokidar would just create separate watchers for the different glob parents resulting in no performance impact.

src/watch/utils.ts
kingston: Without normalization, Windows style separators break the glob logic. Normalizing the paths make all our paths we use consistent and easier to handle (e.g. we can match our globs based on Unix-style separators instead of having to figure out which it is)

kingston (commented):
Sorry for the delay - had a bit of a busy past month.

kingston:
@privatenumber happy new year! just bumping this in case it was still relevant for you.

kingston:
@privatenumber do you think we would be able to merge this pull request at some point? I can resolve conflicts if this seems like a viable candidate to get merged in.

Issue #647, opened 2024-09-05
Request to extend `esbuild` semver range


benmccann:
### Acknowledgements

- [X] I read the documentation and searched existing issues to avoid duplicates
- [X] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [X] I understand this is a free project and relies on community contributions
- [X] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

https://github.com/privatenumber/tsx/blob/157c3ec6bcf0b0a5e387080576404c00f7fd662f/package.json#L80

### Problem & expected behavior (under 200 words)

`esbuild` is a somewhat heavy library with lots of dependencies. I already have another version of `esbuild` pulled in by Vite. Is there any chance you'd accept a PR to extend the range of allowed versions so that I don't end up with multiple versions?

E.g. like this: https://github.com/antfu-collective/importx/blob/08d68f31b4c4e927d59c99e0adb488aa7be9df54/package.json#L55

### Bugs are expected to be fixed by those affected by it

- [X] I'm interested in working on this issue

### Compensating engineering work will speed up resolution and support the project

- [ ] I'm willing to offer $10 for financial support

privatenumber:
Thanks for the issue.

Regarding the impact of the problem, I think this is low priority as users can address it on their end through [dependency overrides](https://pnpm.io/package_json#pnpmoverrides). And most tsx users aren't concerned with dependency de-duplication, as the target audience is broad, extending to beginners.

While dependency size is a valid concern in the npm ecosystem, this issue not unique to tsx as many tools leverage esbuild (even if tsx addresses it, you'll still encounter it from another package). Most users can mitigate this by using pnpm, which minimizes disk usage at a system level. But if you're particularly concerned about esbuild's size, addressing it with esbuild level would likely yield bigger wins. For instance, perhaps splitting the Transform API from the Build API could reduce its size, making a more substantial impact across the ecosystem.

As for the request to extend esbuild's version range in tsx, I'm not keen on the maintenance work involved, including adding/removing versions and reviewing PRs. It also raises process questions: when to add/remove versions, what to do if tsx relies on a new esbuild feature (is it a breaking change to drop old esbuild verisons?), or how to handle compatibility changes (do I need to test old esbuild versions?). I don't mind doing meaningful work for free, but the tradeoff between the maintenance work vs the wins here are questionable.

[...]

PR #664, opened 2024-10-09
feat: upgrade esbuild to 0.24

LinusU:
This bumps esbuild to version 0.24.0.

I wanted this since I'm affected by https://github.com/evanw/esbuild/issues/3887, which prints a lot of warnings when I build with 0.23.0. I also want to use the exact same version of esbuild when running locally, and building for deployment.

Thanks for an amazing tool! 🙏

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
 		"node": ">=18.0.0"
 	},
 	"dependencies": {
-		"esbuild": "~0.23.0",
+		"esbuild": "~0.24.0",
 		"get-tsconfig": "^4.7.5"
 	},
 	"optionalDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       esbuild:
-        specifier: ~0.23.0
-        version: 0.23.0
+        specifier: ~0.24.0
+        version: 0.24.0
       get-tsconfig:
         specifier: ^4.7.5
         version: 4.7.5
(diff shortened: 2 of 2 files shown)


privatenumber:
Thank you for the PR, but I typically don't accept esbuild upgrade requests unless they address a specific bug or issue. In this case, esbuild was released with a [breaking change](https://github.com/evanw/esbuild/releases/tag/v0.24.0) label, which poses unnecessary risk to tsx users without clear benefit.

If your main motivation is to standardize on a single version of esbuild, I recommend using a package manager override to achieve that in your project.

Since tsx is designed to be user-friendly, the average user won't be concerned with dependency de-duplication, and we want to avoid causing potential disruptions unless absolutely necessary.

LinusU:
Makes sense 👍 

I'll look into using overrides instead!

It would be neat if tsx used peer-dependencies to let users bring their own version of esbuild while at the same time indicating which versions are compatible. But as you explained, that probably doesn't vibe with the intention of this project either 😅

theoludwig:
@privatenumber So is that issue #647 possible to be resolved with the goals of tsx (user-friendly)?
There should be a way to avoid duplication of dependencies versions.

Would you accept a PR which change the semver to esbuild from:
`"esbuild": "~0.23.0",`
to
`"esbuild": "~0.23.0 || ~0.24.0",`

or like there: https://github.com/antfu-collective/importx/blob/08d68f31b4c4e927d59c99e0adb488aa7be9df54/package.json#L55
`"esbuild": "^0.20.2 || ^0.21.0 || ^0.22.0 || ^0.23.0",`

PR #669, opened 2024-10-23
fix: upgrade esbuild to 0.24.0 address various issues

aquajach:
Thanks for maintaining this project!

Read and understood the reason behind closing [another PR ](https://github.com/privatenumber/tsx/pull/664). But I would still like to submit this one because:

1. To address [multiple vulnerability issues](https://github.com/evanw/esbuild/pull/3902#issuecomment-2340483728) that from golang. Of course, we could override the version of `esbuild` in our own `package.json` to work around them, but user experience suffers, especially have impact to new users of `tsx` who also care about vulnerabilities.  
2. The `esbuild` release is no longer with `breaking change` label.
3. To avoid a potential `esbuild-wasm` crash coming from a bug in `memchr` https://github.com/evanw/esbuild/pull/3902
4. To reduce the warnings mentioned in [the earlier PR](https://github.com/privatenumber/tsx/pull/664).
5. To fix class field decorators in TypeScript if useDefineForClassFields is false (https://github.com/evanw/esbuild/issues/3913)

To be honest, none of them alone solves a critical problem of `tsx` itself but I hope with all together makes the impact bigger. 

We all understand it is a good practice to keep dependencies up to date to a certain level with cautiousness. Comparing to big version jumps, I think smaller incremental hips is less risky for long run, although brings a bit of more work to do from maintainers' point of view.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
 		"node": ">=18.0.0"
 	},
 	"dependencies": {
-		"esbuild": "~0.23.0",
+		"esbuild": "~0.24.0",
 		"get-tsconfig": "^4.7.5"
 	},
 	"optionalDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       esbuild:
-        specifier: ~0.23.0
-        version: 0.23.0
+        specifier: ~0.24.0
+        version: 0.24.0
       get-tsconfig:
         specifier: ^4.7.5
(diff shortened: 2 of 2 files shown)


privatenumber:
It's marked as a breaking change here:
https://github.com/evanw/esbuild/releases/tag/v0.24.0

Unless you can give me a real reason for upgrading than "good practice to stay up to date", I'll be closing these PRs as it creates unnecessary work for me and has no impact on users.

Closing as a duplicate of https://github.com/privatenumber/tsx/pull/664

Issue #661, opened 2024-10-04
Inspect with chrome as debugger result in ERR_FILE_NOT_FOUND


TzeroOcne:
### Acknowledgements

- [X] I read the documentation and searched existing issues to avoid duplicates
- [X] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [X] I understand this is a free project and relies on community contributions
- [X] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

https://github.com/TzeroOcne/tsx-issue

### Problem & expected behavior (under 200 words)

- What happened
I try to run `tsx` with `--inspect` by `pnpm start:tsx`
- What I expected
I expect I can open debugger tool from browser and it will show the ts file that are beign debugged like the example when I use `ts-node`
- Error stack trace
The error that I got from the debugger are this
```
Could not load content for file:///C:/Users/tzero/temp/tsx-test/src/C:\Users\tzero\temp\tsx-test\src\index.ts (System error: net::ERR_FILE_NOT_FOUND)
```
- Documentation links
I don't think there is any documentation about inspect using chrome as debugger
- Screenshots
This what I expect by using ts-node by running `pnpm start:ts-node`
![image](https://github.com/user-attachments/assets/6889c3d1-6c43-43b3-9945-e5703ce20ae6)
But I got this with tsx by running `pnpm start:tsx`
![image](https://github.com/user-attachments/assets/1f458539-db5f-4e97-8b0a-d4c3515283d9)

### Bugs are expected to be fixed by those affected by it

- [X] I'm interested in working on this issue

### Compensating engineering work will speed up resolution and support the project

- [ ] I'm willing to offer $10 for financial support

privatenumber:
Thanks for the issue. Feel free to work on a PR!

char8x:
I did some simple tests and found the difference on sourcemap

Take the following code as an example

```ts
// test.ts
setInterval(() => {
  debugger
  console.log("aaa");
}, 2000)
```

After executing the `tsx --inspect ./test.ts` command (tsx v4.19.1) , the sourcemap json is generated as follows

```json
{
  "version": 3,
  "mappings": "AAAA,YAAY,IAAM,CAChB,SACA,QAAQ,IAAI,KAAK,CACnB,EAAG,GAAI",
  "names": [],
  "ignoreList": [],
  "sources": [
[...]

PR #670, opened 2024-10-24
fix: generate sourcesContent when Node.js debugger is enabled

char8x:
This PR attempts to fix the issue described here: #661 

If make changes to `tsx/src/run.ts`, the `--inspect` flag will be confused with the flag of the custom program.

```sh
// run command
./dist/cli.mjs --inspect temp/test.ts --inspect-xxx

// tsx/src/run.ts
// `console.log(argv)`
[ '--inspect', 'temp/test.ts', '--inspect-xxx' ]
```

But in `src/utils/transform/get-esbuild-options.ts`, `process.execArgv` doesn't have this problem

```sh
// run command
./dist/cli.mjs --inspect temp/test.ts --inspect-xxx

// src/utils/transform/get-esbuild-options.ts
// `console.log(process.execArgv)`
[
  '--require',
  '/home/debian/tsx/dist/preflight.cjs',
  '--import',
  'file:///home/debian/tsx/dist/loader.mjs',
  '--inspect'
]
```

### DIFF
--- a/src/utils/transform/get-esbuild-options.ts
+++ b/src/utils/transform/get-esbuild-options.ts
@@ -10,12 +10,11 @@ export const baseConfig = Object.freeze({
 	loader: 'default',
 });
 
-const isNodeDebuggerEnabled = process.execArgv.some((flag) => {
+// match Node.js debugger flags
+// https://nodejs.org/api/cli.html#--inspecthostport
+const NODE_DEBUGGER_FLAG_REGEX = /^--inspect(?:-brk|-port|-publish-uid|-wait)?(?:=|$)/;
 
-	// match Node.js debugger flags
-	// https://nodejs.org/api/cli.html#--inspecthostport
-	return /--inspect(-brk|-port|-publish-uid|-wait)?/.test(flag)
-})
+const isNodeDebuggerEnabled = process.execArgv.some(flag => NODE_DEBUGGER_FLAG_REGEX.test(flag));
 
 export const cacheConfig = {
 	...baseConfig,
--- a/src/utils/transform/get-esbuild-options.ts
+++ b/src/utils/transform/get-esbuild-options.ts
@@ -10,6 +10,13 @@ export const baseConfig = Object.freeze({
 	loader: 'default',
 });
 
+const isNodeDebuggerEnabled = process.execArgv.some((flag) => {
+
+	// match Node.js debugger flags
+	// https://nodejs.org/api/cli.html#--inspecthostport
+	return /--inspect(-brk|-port|-publish-uid|-wait)?/.test(flag)
+})
+
 export const cacheConfig = {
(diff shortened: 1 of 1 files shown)


src/utils/transform/get-esbuild-options.ts
```diff
@@ -10,18 +10,25 @@ export const baseConfig = Object.freeze({
 	loader: 'default',
 });
 
+const isNodeDebuggerEnabled = process.execArgv.some((flag) => {
+
+	// match Node.js debugger flags
+	// https://nodejs.org/api/cli.html#--inspecthostport
+	return /--inspect(-brk|-port|-publish-uid|-wait)?/.test(flag)
```
privatenumber, line 17: Can you hardcode the combinations instead?

Regex is slow and hard coding it will be more readable/searchable

char8x: Should I go like this

```js
flag.startsWith('--inspect-brk') || flag.startsWith('--inspect-port') || flag.startsWith('--inspect-publish-uid') || flag.startsWith('--inspect-wait') || flag.startsWith('--inspect')
```

or just this

```js
flag.startsWith('--inspect')
```

Because these are both valid parameters

```sh
tsx --inspect test.ts
tsx --inspect=9229 test.ts
tsx --inspect=127.0.0.1:9229 test.ts

tsx --inspect-brk test.ts
tsx --inspect-brk=9229 test.ts
tsx --inspect-brk=127.0.0.1:9229 test.ts
```

privatenumber: Ah I see, I guess it's more complicated than I expected. Thanks for illustrating the problem.

Let's stick to a regex but can you move it out of the loop so it doesn't get recreated every time, and do something like this instead:
`^--inspect(-brk|-port|-publish-uid|-wait)(=|$)`

char8x: I end up with `/^--inspect(?:-brk|-port|-publish-uid|-wait)?(?:=|$)/`

Because the capturing group is not used, `(?:...)` is used to solve lint errors like `Capturing group number 1 is defined but never used`

privatenumber:
Since we can't really add tests for this, can you confirm this fixes your problem?

I published a prerelease here:

```
npm i 'privatenumber/tsx#npm/fix-node-debugger'
```

or 

```
npx 'privatenumber/tsx#npm/fix-node-debugger' file.ts # same as `npx tsx file.ts`
```

char8x:
I checked it out and it works.

privatenumber:
Thanks for the PR @char8x !

privatenumber:
This issue is now resolved in [v4.19.2](https://github.com/privatenumber/tsx/releases/tag/v4.19.2).

Issue #574, opened 2024-06-06
Support `customConditions` in `tsconfig.json`


colinhacks:
### Acknowledgements

- [X] I read the documentation and searched existing issues to avoid duplicates
- [X] I am reporting a confirmed bug (not seek debugging help)
- [X] I understand this is a collaborative project and relies on my contributions
- [X] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/develop/CONTRIBUTING.md)

### Minimal reproduction URL

https://github.com/colinhacks/live-typescript-monorepo

### Problem & expected behavior (under 200 words)

```
gh repo clone https://github.com/colinhacks/live-typescript-monorepo
cd live-typescript-monorepo/custom-conditions
pnpm i
cd packages/pkg-b
tsx index.ts # module resolution error
tsx --conditions source index.ts # works!
```

### Bugs are expected to be fixed by those affected by it

- [ ] I'm interested in working on this issue

### Compensating engineering work financially will speed up resolution

- [ ] I'm willing to offer financial support

privatenumber:
Feature requests are reserved for sponsors, but I'll accept this for you @colinhacks!

This should be trivial to support for ESM, but it may be challenging with the CJS resolver as it doesn't seem to offer a way to accept conditions:
https://github.com/nodejs/node/blob/main/lib/internal/modules/cjs/loader.js#L1126

DASPRiD:
I just ran into this as well in a mono repo. For now you can work around this with the following command line instead:

`tsx -C development your-script.ts`

privatenumber:
Looks like Node added customConditions in the CommonJS loader: https://github.com/nodejs/node/pull/55698/files#diff-5b5902273122e094ff474fda358605ffa45a4a58b51cd0bf4c1acb93779df142R1326

This may be feasible to implement in tsx now.

PR #675, opened 2024-11-17
feat(esm): support custom conditions

colinlienard:
Add support for [customConditions](https://www.typescriptlang.org/tsconfig/#customConditions) in ESM.

For example, if a tsconfig contains this:

```json
{
	"compilerOptions": {
		"customConditions": ["source"],
	}
}
```

passing the condition as an arg to the CLI would not be needed anymore:

```diff
- tsx --conditions source src
+ tsx src
```

But this is only for ESM as explained here: https://github.com/privatenumber/tsx/issues/574#issuecomment-2156442343

Related to #574


privatenumber:
Thanks for the PR

This is blocked until we add condition support to the CJS resolver though. Since tsx aims to consolidate ESM and CJS, we want to avoid introducing discrepancies between them.

PR #698, opened 2025-02-11
fix: upgrade `esbuild` to `~0.25.0` to address vuln report

RobinTail:
Fixes [GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-67mh-4wv8-2f99)

https://github.com/evanw/esbuild/releases/tag/v0.25.0

I ran the tests manually and they succeeded.

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
 		"node": ">=18.0.0"
 	},
 	"dependencies": {
-		"esbuild": "~0.23.0",
+		"esbuild": "~0.25.0",
 		"get-tsconfig": "^4.7.5"
 	},
 	"optionalDependencies": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -9,8 +9,8 @@ importers:
   .:
     dependencies:
       esbuild:
-        specifier: ~0.23.0
-        version: 0.23.0
+        specifier: ~0.25.0
+        version: 0.25.0
       get-tsconfig:
         specifier: ^4.7.5
         version: 4.7.5
@@ -288,6 +288,12 @@ packages:
     cpu: [ppc64]
     os: [aix]
 
+  '@esbuild/aix-ppc64@0.25.0':
+    resolution: {integrity: sha512-O7vun9Sf8DFjH2UtqK8Ku3LkquL9SZL8OLY1T5NZkA34+wG3OQF7cl4Ql8vdNzM6fzBbYfLaiRLIOZ+2FOCgBQ==}
+    engines: {node: '>=18'}
+    cpu: [ppc64]
+    os: [aix]
+
   '@esbuild/android-arm64@0.21.5':
     resolution: {integrity: sha512-c0uX9VAUBQ7dTDCjq+wdyGLowMdtR/GoC2U5IYk/7D1H1JYC0qseD7+11iMP2mRLN9RcCMRcjC4YMclCzGwS/A==}
     engines: {node: '>=12'}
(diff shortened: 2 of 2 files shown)


RobinTail:
@privatenumber , please review

LukeNotable:
How about specifying a broader range like `>= 0.22.0 < 0.26.0` to support the latest without requiring everyone to upgrade through potentially breaking changes (if that's a concern)?

casewalker:
@LukeNotable Sadly, that's kind of out of tsx's hands. Semver [breaks down for versions < 1.0.0](https://semver.org/#spec-item-4), and even though it is a little more than 5 years old, esbuild has not yet decided to go from  from versions 0.y.z to an official released 1.0.0 version yet. Since there are no guarantees pre-1.0.0, tsx cannot trust that there won't be breaking changes even in the next patch version, i.e. 0.25.1. I find that pretty frustrating, so I just [left a comment over there](https://github.com/evanw/esbuild/issues/2544#issuecomment-2655360075) asking that esbuild reconsider their versioning strategy.

tats-u:
Even though tsx itself isn't affected by this vulnerability, no one can imagine the culprit that emitted security alerts in projects is tsx.

tats-u:
Changes in 0.24

https://github.com/evanw/esbuild/releases/tag/v0.24.0

None of them or 0.25 affect on tsx, right?

shinebayar-g:
> tsx cannot trust that there won't be breaking changes even in the next patch version, i.e. 0.25.1.

Small correction. Before 1.0.0, Major = Minor. So patch versions _should NOT_ introduce breaking changes.

casewalker:
@shinebayar-g No, that's not how SemVer is defined. If you clicked the [link I shared before](https://semver.org/#spec-item-4), it says:
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time.

It doesn't say breaking changes are restricted to only minor version increases, it says anything may change at any time.

It may be true that _some projects_ treat the Minor and Patch versions in this down-shifted way which you describe, and I agree it is super helpful when they do, but SemVer does not guarantee that.

shinebayar-g:
Fair.

luhn:
> tsx cannot trust that there won't be breaking changes even in the next patch version, i.e. 0.25.1

But tsx is *already* trusting this, it currently depends on esbuild `~0.23.0` which includes `0.23.1` and any possible future patch versions.

image

tats-u:
0.24.x or 0.25.x are not included there. 
Why do you think esbuild provides a patch for 0.23.x only in favor of tsx?

luhn:
My point is that tsx is already assuming that patch releases are backwards compatible, so if `0.25.0` passes muster then `<0.26` is a valid constraint

tats-u:
<0.26 allows vulnerable versions (0.24.x or below), too. 
The current content in this PR is the best.

tats-u:
It depends on the packages. However esbuild doesn't seem to contaminate patch updates with breaking changes.
The current content of this PR is the best.

tats-u:
It depends on the packages. However esbuild doesn't seem to contaminate patch updates with breaking changes.
The current content of this PR is the best.

tats-u:
It depends on the packages. However esbuild doesn't seem to contaminate patch updates with breaking changes.
The current content of this PR is the best.

If someone insists that esbuild brings breaking changes even to patch updates, they must paste the change log of one of versions other than 0.x.0 that included breaking changes here.

naile:
Are there plans on merging this soon? Would be nice to get rid of this from our vulnerability reports.

privatenumber:
This issue is now resolved in [v4.19.3](https://github.com/privatenumber/tsx/releases/tag/v4.19.3).

PR #710, opened 2025-03-21
fix: support nyc coverage discovery

### LINKED ISSUE #699: tsx issue with nyc instumented files:
### Acknowledgements

### Minimal reproduction URL

https://stackblitz.com/edit/node-r2m8mrnc?file=package.json

- What happened

  When using `nyc` and `tsx` the instumented files cause an error:
  `Cannot read properties of undefined (reading 'startsWith')`.

  I think this line is causing the issue, as modules don't always have an id:
  https://github.com/privatenumber/tsx/blob/28a3e7d2b8fd72b683aab8a98dd1fcee4624e4cb/src/cjs/api/module-extensions.ts#L107

- How to repoduce
  
  Use stackblitz or manual steps:
  ```bash
  npm init -y
  npm i tsx@4.19.2 nyc@17.1.0 --save-dev --save-exact
  echo "node-options=--import=tsx" > .npmrc
  echo "{\"all\": true, \"sourceMap\": true, \"instrument\": true}" > .nycrc.json
  echo "console.log('hello world')" > dummy.js
  npx nyc node dummy.js
  ```

- Screenshots
  Image
  Image

derdeka:
## Problem

nyc's `all: true` coverage discovery asks the active CommonJS extension loader to transform each candidate file before running the program. It supplies a minimal module-like object with `_compile` but no `id`.

tsx assumed every loader input had an `id` and called `startsWith()` unconditionally. Projects that preload tsx with nyc therefore crashed with `Cannot read properties of undefined (reading 'startsWith')` before application code ran.

## Changes

tsx now applies its data-URL-specific `module.path` correction only when the loader input provides a matching ID. Regular CommonJS modules retain their existing behavior, while nyc's source-discovery module reaches `_compile` so coverage collection can continue.

A regression test invokes the extension with nyc's minimal module shape and asserts that it receives the fixture source.


privatenumber:
This issue is now resolved in [v4.23.10](https://github.com/privatenumber/tsx/releases/tag/v4.23.10).

PR #712, opened 2025-04-02
feat(watch): prefer includes over excludes

### LINKED ISSUE #210: Support for watching hidden files
### Feature request

Allows monitoring changes starting with `.`, like `.env` or other hidden files.

### Why?

Reading hidden files in ts files is a special requirement we encounter in business development. 

For example, we will keep some secret data locally on the server by means of hidden files. When adjusting the content of hidden files, the result should also be real-time refresh, but tsx excludes hidden files from the monitoring range by default, which will not meet this requirement. 

For the same case, please refer to [issue774](https://github.com/esbuild-kit/tsx/issues/181)

### Alternatives

In order to meet such requirements, it is necessary to add the monitoring capability of dependent files other than js/ts first, like the capabilities provided by https://github.com/esbuild-kit/tsx/pull/208.

Then exclude rules that match the currently watched file from the default ignored rules. For example, if I watch `.env` file,  `**/.*` should be removed from the ignored rules, as shown in the figure below.

image

### Additional context

_No response_

mogzol:
Fixes #210

This change makes any explicitly included files in watch mode never get excluded, even if they otherwise would be. Particularly, this is useful for watching files that are excluded by the default excludes, like ".env" files, which were previously impossible to watch.

In making these changes, chokidar has also been upgraded to v4, which no longer supports glob patterns. Glob functionality is now achieved with the 'glob' library. The `windowsPathsNoEscape` glob option is enabled in order to keep the behavior the same as it was with chokidar v3, where backslash cannot be used as a glob escape character, and only as a directory separator.

### Example

Consider the command:

```
tsx watch --exclude="./lib" --include="./lib/special.ts" index.ts
```

Previously, changes to `./lib/special.ts` would not trigger a restart, despite the fact that it was explicitly watched, because it's containing folder was excluded. With these changes, a restart will be triggered, but changes to any other files in `./lib` will still be ignored.

This is also true for included globs. For example, this command:

```
tsx watch --exclude="./lib" --include="./lib/dir/**" index.ts
```

Will restart if any files under `./lib/dir` are modified, but will not restart for any other changes in the lib directory.


mogzol:
I rebased this onto the latest commit to resolve conflicts

mogzol:
Updated this to fix a minor issue where the `windowsPathsNoEscape` option wasn't passed to `hasMagic()`. Also removed the redundant `**/.*` default ignore pattern, since it is already covered by the `**/.*/**` pattern (a glob ending in `/**` also matches the path without the `/**`)

ragrag:
@privatenumber mind taking a look at this? 🙏🏻 
it's been a long standing issue and a very useful feature since currently it's not possible to watch files like .env trivially without patching tsx

PR #733, opened 2025-08-11
fix: ignore native typescript formats

marco-ippolito:
Fixes: https://github.com/nodejs/node/issues/59364
This patch fixes compatibility with Node v24 and v22 type stripping.


privatenumber:
Thanks for the patch!

FYI we don't yet claim full/official support for the latest Node behavior around type stripping, `require(esm)`, and all the other improvements. tsx leans on Node internals in a few places, and with recent changes there's a lot of catching up for us to do.

That said, this change is small, targeted, and improves compatibility right away, so I'm happy to merge it. I'll get this in and cut a release shortly.

Issue #163, opened 2022-12-12
Can't use cli in watch mode


grind-t:
### Bug description

What did you do? (Provide reproduction code in next section)
Run `npx tsx watch` for console app

What did you expect to happen?
I was expecting the ability to enter text into the console

What happened instead?
Entering text into the console starts rerun

Do you have an error stack-trace or screenshots?
![image](https://user-images.githubusercontent.com/49301275/207144913-68ff90d3-78e9-4654-83cf-b7678022114e.png)

### Reproduction

 

index.ts:
```ts
import { createInterface } from 'readline'

const rl = createInterface({
  input: process.stdin,
  output: process.stdout,
})

rl.on('line', async (line) => {
  console.log(line)
  rl.prompt()
})
```

command: `npx tsx watch --clear-screen=false ./index.ts`

### Environment

```shell
<!--
  Run and paste the output of:
  
  npx envinfo --system --npmPackages tsx --binaries
  ```
-->

System:
  OS: Linux 5.15 Ubuntu 22.04.1 LTS 22.04.1 LTS (Jammy Jellyfish)
  CPU: (8) x64 Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz
  Shell: 5.1.16 - /bin/bash
Binaries:
  Node: v18.12.1
  npm: 
npmPackages:
  tsx: ^3.12.1 => 3.12.1
```


### Can you contribute a fix?

- [ ] I’m interested in opening a pull request for this issue.

privatenumber:
We can fix this in a breaking change release by removing https://github.com/esbuild-kit/tsx/blob/be0478990305fc0052d9c6c982a20d3990658cc2/src/watch/index.ts#L166

but tsx would need another way to manually reload watch.

Will have to look into how other watchers handle this.

ben12:
> We can fix this in a breaking change release by removing
> 
> https://github.com/esbuild-kit/tsx/blob/be0478990305fc0052d9c6c982a20d3990658cc2/src/watch/index.ts#L166
> 
> but tsx would need another way to manually reload watch.
> 
> Will have to look into how other watchers handle this.

To avoid breaking change, a command line option to disable this feature could be added. `--no-keyboard-rerun` for example ?

tizmagik:
I’m having the same issue. Would the extra command line option be acceptable? If so, I can work on a PR

privatenumber:
[...]

PR #753, opened 2025-11-05
feat: add flag to disable tsx watch from reading from process.stdin

kumavis:
### Acknowledgements

### bug

There are som limitations to an inherited stdin:
- When the child process reads from stdin, the parent process also reading from stdin causes conflicts (race to read, only one reads).
- When the child process sets the stdin mode to "raw mode", every key press seen by the parent (see race above) causes the parent to restart the child.

### proposal

Add an option to disable the stdin reads for the reload check

The default was selected so as to not introduce a breaking change.
The flag name `--reload-on-keypress`(a bit inexact), was chosen to leave room for the reload triggering key to change via a future flag.

### Relevant issues:
- essentially the same PR https://github.com/privatenumber/tsx/pull/220/files
- help wanted: alternatives for triggering reload https://github.com/privatenumber/tsx/issues/215#issuecomment-1794274016
- watch breaks stdin for child https://github.com/privatenumber/tsx/issues/163
- `process.stdin.isRaw` does not necessarily reflect the actual mode https://github.com/nodejs/node/issues/47938

### other notes
the child sharing a stdin adds some complications, as the child may expect a certain mode or change the mode. its possible to instead do set stdin to `'pipe'` in the spawn config, set the parent in raw mode, and pipe input into the child. another alternative may be to give the child a pseudoterminal https://github.com/microsoft/node-pty
i haven't explored these options in depth

### DIFF
--- a/docs/watch-mode.md
+++ b/docs/watch-mode.md
@@ -51,3 +51,4 @@ tsx watch --exclude "./data/**/*" ./file.ts
 
 - Press <kbd>Return</kbd> to manually rerun the script.
 - Use `--clear-screen=false` to prevent the screen from clearing on rerun.
+- Use `--reload-on-keypress=false` to disable reloading on keypress. If your child process uses stdin raw mode, be sure to disable this.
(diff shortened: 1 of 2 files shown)


Jodu555:
Hey! This change would be super helpful for interactive CLI use cases since pressing enter instantly reloads
Would love to see it merged when you get a chance!

PR #768, opened 2026-01-16
fix(esm/hook/resolve): tsconfig when in node_modules

thisago:
(It seems that some tests was already failing)

This removes the check if any of parent directory is `node_modules`,
allowing [a lib to be built manually, in cases it is installed directly
from a Git repo](https://browse.library.kiwix.org/content/stackoverflow.com_en_all_2023-11/questions/48287776/automatically-build-npm-module-on-install-from-github)

Example reproducing the issue https://github.com/thisago/tsx-issue-importing-from-node-modules

It seems to be [intended because](https://github.com/thisago/tsx/commit/e46366d2308afdf2dd197165854ff48f94a4b753) [it had tests](https://github.com/privatenumber/tsx/pull/360), but this behavior makes impossible to run a file when it's inside a `node_modules`.

### DIFF
--- a/src/esm/hook/resolve.ts
+++ b/src/esm/hook/resolve.ts
@@ -249,14 +249,12 @@ const resolveTsPaths: ResolveHook = async (
 
 		requestAcceptsQuery: requestAcceptsQuery(specifier),
 		tsconfigPathsMatcher,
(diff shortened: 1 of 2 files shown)


thisago:
Hello @privatenumber, any possibility of this get merged?

Issue #469, opened 2024-01-31
Configurable timeout in watch before force kill


jonasb:
### Precheck

- [X] I searched existing issues before opening this one to avoid duplicates
- [X] I understand tsx aims for TypeScript parity and will not support arbitrary Node.js enhancements
- [X] This request cannot be made to Node.js directly, and is specific to tsx
- [X] I understand this form is not for addressing a bug or seeking help

### Feature request

I would like to configure the 5s timeout before the process in force killed (`Process didn't exit in 5s. Force killing...`) when I've made changes in watch mode.

Either environment variable or a command line switch would do the job. If it would be possible to turn off the timeout (e.g. with 0 or a negative timeout), that would be nice.

### Motivations

I have long running (but not infinite) requests running in my web server (up to a minute or so). I'd like for those to finish before the process is killed and restarted.

### Contributions

- [ ] I plan to open a pull request for this issue
- [ ] I plan to make a financial contribution to this project

zachkirsch:
Would love this too!

PR #769, opened 2026-01-18
feat: Make kill-timeout configurable

CommanderRedYT:
### Why I thought about this
My application registers custom handlers with calls like `process.on('SIGINT')`. In there, I am waiting to gracefully shutdown another process. This takes some time, sometimes more than 5 seconds. The idea is to have this configurable so that no `SIGKILL` is sent that breaks saving in the childprocess.

### Description
This PR implements a new flag `--kill-timeout <number>` for the `watch` subcommand. The number is used in the timeout to change and possibly extend the grace period until the process is killed after sending the SIGINT.

I am happy to implement any feedback provided.

### DIFF
--- a/src/watch/index.ts
+++ b/src/watch/index.ts
@@ -32,6 +32,11 @@ const flags = {
 		description: 'Clearing the screen on rerun',
 		default: true,
 	},
+	killTimeout: {
+		type: Number,
+		description: 'Milliseconds until process is forcibly killed'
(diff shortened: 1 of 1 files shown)


niss36:
As mentioned in https://github.com/privatenumber/tsx/issues/469 this would be super helpful 🙏

zachkirsch:
+1

Issue #767, opened 2026-01-12
tsx watch doesn't detect file changes on node 24.x


Germain-condo-group:
### Acknowledgements

- [x] I read the documentation and searched existing issues to avoid duplicates
- [x] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [x] I understand this is a free project and relies on community contributions
- [x] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

none

### Problem & expected behavior (under 200 words)

After upgrading from node v22.21.1 to node v24.12.0, tsx watch mode stopped detecting file changes.
Previously, it would restart app upon file change, but after upgrading node to 24.12.0 it never restarts.

tsx version: `4.21.0`

The watch command we use is the following:
`TZ=utc tsx watch --inspect=9229 --include .env --env-file=.env --conditions=development src/index.ts -c dev`

Our index.ts file
```ts
import { PgInstrumentation } from '@opentelemetry/instrumentation-pg';
import './dsn';
import { parseEnv } from './env';
import 'pgvector';

const env = parseEnv();
// this must run before any app-code as it augments modules
const pgInstrumentation = new PgInstrumentation({
    // add query + params to span attributes
    enhancedDatabaseReporting: true,
    requireParentSpan: true,
});

void import('./app.js').then(mod => mod.createApp(env, [pgInstrumentation]).run());
```

### Bugs are expected to be fixed by those affected by it

- [ ] I'm interested in working on this issue

### Compensating engineering work will speed up resolution and support the project

- [ ] I'm willing to offer $10 for financial support

luastoned:
Since Node v23 require ESM might take a different (native?) code path.
This causes `parent.send({ type: 'dependency', path: url });` to not be called to register the file with the watcher.

Changing `"type": "module"` in `package.json` fixes this temporarily.

stephenh:
Fwiw @privatenumber afaict `watch` was working just fine in `v24.11.0` (and I assume prior--we'd been using `v24.x` releases like `v24.9.0` for awhile with `tsx watch working).

But now `watch` is broken in the latest `v24.12.0` and `v24.13.0` releases.

Here's the changelog of v24.12.0:

https://nodejs.org/en/blog/release/v24.12.0

Looks like maybe there are some internal `watch` changes?

SamuelGaona:
Having the same problem, workaround cited by @luastoned works well.

[...]

PR #774, opened 2026-02-02
store IPC messages before connection

giuseppelt:
In the new node v24, I found out the IPC messages are lost as the IPC connection is yet to be setup/connected.
This is maybe due to a different import/load order strategy in node.

Anyway, a simple fix was to just cache the messages and replay them when the connection is established.

This fixes the "watch doesn't restart on changes".
Solves #767

### DIFF
--- a/src/utils/ipc/client.ts
+++ b/src/utils/ipc/client.ts
@@ -35,8 +35,10 @@ const connectToServer = () => new Promise<SendToParent | void>((resolve) => {
 	socket.unref();
 });
 
+// before connection, store messages locally
+const queue = [];
 export const parent: Parent = {
-	send: undefined,
+	send: x => queue.push(x),
 };
 
 export const connectingToServer = connectToServer();
@@ -44,6 +46,10 @@ export const connectingToServer = connectToServer();
 connectingToServer.then(
 	(send) => {
 		parent.send = send;
+		
+		// process the queue
+		queue.forEach(x => send(x));
+		queue.length = 0;
 	},
 	() => {},
 );
--- a/src/utils/ipc/client.ts
+++ b/src/utils/ipc/client.ts
@@ -48,8 +48,10 @@ connectingToServer.then(
 		parent.send = send;
 		
 		// process the queue
-		queue.forEach(x => send(x));
-		queue.length = 0;
+        if (!!send) {
+		  queue.forEach(x => send(x));
+		  queue.length = 0;
+		}
 	},
 	() => {},
 );


intelmib:
up

src/utils/ipc/client.ts
```diff
@@ -35,15 +35,21 @@ const connectToServer = () => new Promise<SendToParent | void>((resolve) => {
...
 
 export const connectingToServer = connectToServer();
 
 connectingToServer.then(
 	(send) => {
 		parent.send = send;
+		
+		// process the queue
+		queue.forEach(x => send(x));
+		queue.length = 0;
```
santialbo: https://github.com/privatenumber/tsx/blob/938e46c21e2f1a5d654f8b7fb0d0a9a6390834ea/src/utils/ipc/client.ts#L31

send could be undefined


```suggestion
        if (!!send) {
		  queue.forEach(x => send(x));
		  queue.length = 0;
		}
```

(I'm not a mantainer, I just faced this issue after applying the patch on my codebase)

privatenumber:
Possible to add a failing test to cover this?

giuseppelt:
> Possible to add a failing test to cover this?

To be honest, I don't know where to start. 

How to reproduce the node loader behavior for the specific version in an independent harness, I don't know. And as this fix is just a "null-check", the test (and its maintenance) would be way more complex than the fix itself.

intelmib:
While waiting for #774 to be merged, here's a post-install patch script that applies the same queue-based fix directly to the compiled dist files (client-*.mjs / client-*.cjs). Useful for anyone stuck on Node 24 in production.

It's idempotent — skips if already patched, creates backups, and handles both ESM and CJS variants.

📎 https://gist.github.com/intelmib/091dd8fd0684913f40babd04418dfa5a

privatenumber:
Thanks for narrowing this down. This diagnosis was correct: dependency IPC messages could be emitted before the IPC connection was ready, which caused `tsx watch` to miss runtime dependencies on newer Node loader paths.

This is now fixed in [v4.21.1](https://github.com/privatenumber/tsx/releases/tag/v4.21.1) by [`1d7e528`](https://github.com/privatenumber/tsx/commit/1d7e528762a7e4f801175fd7d7d6082b00df3e5c). The released fix keeps the queueing behavior from this PR, adds cleanup for the no-server/error path, and includes version-sensitive watch coverage for the sync-hook Node path.

Closing this as superseded by the released fix. Appreciate the investigation here.

PR #784, opened 2026-04-13
fix: queue IPC messages sent before socket connection is established

### LINKED ISSUE #767: tsx watch doesn't detect file changes on node 24.x

marcus13371337:
Implements the fix from #774 with an added regression test.

I noticed #774 has been open for a while and I can't push to someone else's PR, so I'm opening this one with the same fix plus a test to cover it.

## What's broken

On Node v24, CJS modules load fast enough that `parent.send()` is called (from the require hook) before the IPC socket connection to the watch server is established. Since `parent.send` was initialised to `undefined`, the `if (parent.send)` guard in the hook silently drops every dependency message — so the watcher never learns about imported files and watch mode never restarts on changes.

## Fix

Initialise `parent.send` to a queue function instead of `undefined`. Messages sent before the socket connects are buffered, then flushed once the connection resolves. If the connection fails (no watch server), the queue is discarded.

## Test

Added `tracks dependencies loaded before IPC connects` to `tests/specs/watch.ts`:
- Uses a `commonjs` package fixture (the package type where the race manifests on Node 24)
- Changes the imported dependency immediately after first output — no artificial delay
- **Fails on Node 24.14.1 without the fix** (times out waiting for restart)
- **Passes on Node 24.14.1 with the fix**
- Passes on Node 20 regardless (timing doesn't trigger the race there)

Closes #767

### DIFF
--- a/src/utils/ipc/client.ts
+++ b/src/utils/ipc/client.ts
@@ -35,8 +35,10 @@ const connectToServer = () => new Promise<SendToParent | void>((resolve) => {
 	socket.unref();
 });
 
+// Buffer messages sent before the connection is established
+const queue: Record<string, unknown>[] = [];
 export const parent: Parent = {
-	send: undefined,
+	send: x => queue.push(x),
 };
 
 export const connectingToServer = connectToServer();
@@ -44,6 +46,10 @@ export const connectingToServer = connectToServer();
 connectingToServer.then(
 	(send) => {
 		parent.send = send;
+		if (send) {
+			queue.forEach(x => send(x));
+			queue.length = 0;
+		}
 	},
 	() => {},
 );
--- a/tests/specs/watch.ts
+++ b/tests/specs/watch.ts
@@ -74,6 +74,55 @@ export default testSuite(async ({ describe }, { tsx }: NodeApis) => {
 			}, 10_000);
 		}
 
+		test('tracks dependencies loaded before IPC connects', async ({ onTestFinish, onTestFail }) => {
+			const fixtureWatch = await createFixture({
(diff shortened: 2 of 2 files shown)


privatenumber:
Thanks for adding the test coverage around the IPC race. This is now fixed in [v4.21.1](https://github.com/privatenumber/tsx/releases/tag/v4.21.1) by [`1d7e528`](https://github.com/privatenumber/tsx/commit/1d7e528762a7e4f801175fd7d7d6082b00df3e5c).

The released implementation keeps the same queueing idea, adds cleanup for the no-server/error path, and includes version-sensitive watch coverage for reruns when imported TypeScript dependencies change. Closing this as superseded by the released fix.

PR #807, opened 2026-06-10
fix: isolate ESM hook state per module.register() registration

### LINKED ISSUE #806: `tsImport()` hangs forever instead of rejecting after ~5 registrations (regression in 4.22.0, Node < 24.11.1)
### Acknowledgements

### Minimal reproduction URL

https://github.com/rexxars/tsx-tsimport-hang

Regression in 4.22.0 (4.21.0 works). On Node versions where tsx uses async `module.register()` (below 24.11.1 / 25.1.0: all of Node 18/20/22/23 and most 24.x), `tsImport()` of a module containing a failing relative import stops settling once enough registrations accumulate: calls 1-5 reject with `ERR_MODULE_NOT_FOUND` as expected, the 6th call never settles and the process hangs forever, idle. Any failing relative import from a TypeScript file triggers it; an unloadable extension like .svg hangs the same way with `ERR_UNKNOWN_FILE_EXTENSION`.

Expected: all calls reject and the script completes, as on 4.21.0 and on Node >= 24.11.1 (the `registerHooks` path).

Cause (debugged): since 4.22.0, `src/esm/api/register.ts` imports the hook factories, so bundling moved the hook module, including the mutable data singleton (namespace, active flag), into a shared chunk. The cache-busted `./esm/index.mjs?<timestamp>` entry re-evaluates per registration, but the shared chunk evaluates once, so every `initialize()` overwrites the same state. All chained registrations then claim the latest namespace, and each re-applies TypeScript extension guessing to the failed candidates of the hook above it, multiplying resolution work roughly 10x per registration. `unregister()` doesn't help since active is shared too.

- [x] I'm interested in working on this issue

rexxars:
- `tsImport()` hangs once a failing import passes through 6 or more hook registrations on Node versions using the async `module.register()` path (below 24.11.1 / 25.1.0). Regressed in 4.22.0.
- The `api/register` module started importing the hook factories, which made the bundler move the hook module, including its mutable hook state singleton, into a shared chunk. The cache-busted loader entry re-evaluates per registration, but the shared chunk only evaluates once, so every `initialize()` overwrote the same namespace and active flag.
- With all chained registrations claiming the most recent namespace, each hook re-applied TypeScript extension guessing to the failed candidates of the hook above it, multiplying resolution work per registration. Failing imports never settled, and unregistering could not deactivate the shared state.

This PR:
- Creates hook state in the loader entry's module scope via a `createHooks()` factory so each registration owns its data, and drops the state-bound singleton exports from the hook modules.
- Suffixes the loader URL query with a counter so same-millisecond registrations cannot reuse the cached loader module and share state.
- Covers repeated failing `tsImport()` calls across registrations in version-sensitive tests.

Fixes #806

### DIFF
--- a/src/esm/api/register.ts
+++ b/src/esm/api/register.ts
@@ -37,6 +37,10 @@ export type Register = {
 
 let cjsInteropApplied = false;
 
+// Deduplicates same-millisecond registrations, which would otherwise reuse
+// the same cached loader module and share per-registration hook state
+let registrationCount = 0;
+
 const collectImportSpecifiers = (
 	argv: string[],
 ) => {
@@ -154,10 +158,12 @@ export const register: Register = (
 		return unregister;
 	}
 
+	registrationCount += 1;
+
 	const { port1, port2 } = new MessageChannel();
 	module.register(
 		// Load new copy of loader so it can be registered multiple times
-		`./esm/index.mjs?${Date.now()}`,
+		`./esm/index.mjs?${Date.now()}.${registrationCount}`,
 		{
 			parentURL: import.meta.url,
 			data: {
--- a/src/esm/hook/index.ts
+++ b/src/esm/hook/index.ts
@@ -1,3 +1,17 @@
-export { initialize, globalPreload } from './initialize.js';
-export { load } from './load.js';
-export { resolve } from './resolve.js';
+import { createInitialize, createGlobalPreload, type Data } from './initialize.js';
+import { createLoad } from './load.js';
+import { createResolve } from './resolve.js';
+
+export const createHooks = () => {
+	const data: Data = {
+		active: true,
+		parsedTsconfig: undefined,
+	};
+
+	return {
+		initialize: createInitialize(data),
+		globalPreload: createGlobalPreload(data),
+		load: createLoad(data),
+		resolve: createResolve(data),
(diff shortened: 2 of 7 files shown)


rexxars:
@cschanhniem good call - added in ed37ebe.

New test `concurrent registrations keep their namespaces isolated`: two registrations stay active at once, each imports the same module, and it asserts each resolves under its own `tsx-namespace` tag (plus a per-namespace load count, so a cross-namespace cache hit would show up too).

One thing worth flagging for whoever touches this next: the test only catches the regression because both registrations are alive at the same time. I tried the obvious sequential version first (register one → import → register two → import) and it passes even against the buggy build: by the time the second `initialize()` runs, the first registration is already done and its hooks are idle, so the clobbering is invisible. Same reason the original bug only surfaced on the error path, which keeps the old registrations chained and active.

Confirmed it fails against the pre-fix build for the right reason: registering the second namespace overwrites the shared one, so the first registration's `tsx://` scoped-import specifier matches no active hook and leaks through to Node (`Cannot find package 'tsx:'`).

Scoped it to the namespaced-vs-namespaced case, since the existing `namespace allows async nested calls without cross contamination` test already covers namespaced-vs-default.

Issue #809, opened 2026-06-22
tsx is way slower since v4.21.1


leia-uwu:
### Acknowledgements

- [x] I read the documentation and searched existing issues to avoid duplicates
- [x] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [x] I understand this is a free project and relies on community contributions
- [x] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

https://github.com/survev/survev/

### Problem & expected behavior (under 200 words)

I recently noticed that tsx has been around 2 times slower on my project, and git bisected it to commit 1d7e528762a7e4f801175fd7d7d6082b00df3e5c

It seems to be something related to the import loader so i can't really provide a "minimum" reproduction URL because its only noticeable on projects that imports large amounts of files

Steps to reproduce:
- clone the project from the reproduction URL ( https://github.com/survev/survev/ )
- Install dependencies with `pnpm install`
- cd into `server` directory
- run `tsx src/gameServer.ts` or `pnpm dev:game` script
- wait for a log that has `Game Process Manager | [INFO] Process {PID} created in {TIME}ms`

Observed behavior:

Since commit 1d7e528762a7e4f801175fd7d7d6082b00df3e5c (and release v4.21.1) it will take way longer to create the game process

On my Arch Linux desktop system it went from around 800ms to almost 2 seconds:

<img width="1003" height="785" alt="Image" src="https://github.com/user-attachments/assets/4be25019-91fc-4cdf-8b64-11daaba6740d" />

And on an M4 Macbook with MacOS it went from around 300ms to around 900ms

Tested with nodejs v24.16.0 on both systems

### Bugs are expected to be fixed by those affected by it

- [ ] I'm interested in working on this issue

### Compensating engineering work will speed up resolution and support the project

- [ ] I'm willing to offer $10 for financial support

mtmacdonald:
We are also seeing 4x or more slower performance and have pinned our repos on `4.21.0` in the meantime. It seems like [this changeset](https://github.com/privatenumber/tsx/commit/1d7e528762a7e4f801175fd7d7d6082b00df3e5c) may be the relevant one.

privatenumber:
This issue is now resolved in [v4.23.0](https://github.com/privatenumber/tsx/releases/tag/v4.23.0).

If you're able to, [your sponsorship](https://github.com/sponsors/privatenumber) would be very much appreciated.

Issue #640, opened 2024-08-21
Fix in 4.16.3 broke use of certain libraries when allowJs is true


lawrenceong:
### Acknowledgements

- [X] I read the documentation and searched existing issues to avoid duplicates
- [X] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [X] I understand this is a free project and relies on community contributions
- [X] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

https://stackblitz.com/edit/node-epykqv

### Problem & expected behavior (under 200 words)

From and including version 4.16.3 of `tsx`, the bug caused the following problem:

If `tsconfig.json` contains a compiler option of `allowJs` set to `true`:

```
{
  "compilerOptions": {
    "allowJs": true
  }
}
```

Then certain packages will fail when used with `tsx`.  From the minimal reproduction URL, using `libphonenumber-js` and `tsx`, a simple call will result in the following error:

```
> tsx index.ts

Error: [libphonenumber-js] `metadata` argument was passed but it's not a valid metadata. Must be an object having `.countries` child object property. Got an object of shape: { default }.
    at validateMetadata (/home/projects/node-epykqv/node_modules/.pnpm/libphonenumber-js@1.11.7/node_modules/libphonenumber-js/build/metadata.js:591:11)
    at new Metadata (/home/projects/node-epykqv/node_modules/.pnpm/libphonenumber-js@1.11.7/node_modules/libphonenumber-js/build/metadata.js:43:5)
    at parse (/home/projects/node-epykqv/node_modules/.pnpm/libphonenumber-js@1.11.7/node_modules/libphonenumber-js/build/parse.js:89:14)
    at parsePhoneNumberWithError (/home/projects/node-epykqv/node_modules/.pnpm/libphonenumber-js@1.11.7/node_modules/libphonenumber-js/build/parsePhoneNumberWithError_.js:19:32)
    at parsePhoneNumber (/home/projects/node-epykqv/node_modules/.pnpm/libphonenumber-js@1.11.7/node_modules/libphonenumber-js/build/parsePhoneNumber_.js:32:55)
    at parsePhoneNumber (/home/projects/node-epykqv/node_modules/.pnpm/libphonenumber-js@1.11.7/node_modules/libphonenumber-js/build/parsePhoneNumber.js:20:44)
[...]

PR #811, opened 2026-06-28
fix: avoid redundant filesystem probes during module resolution

wrose504:
## Overview

Fixes a startup-time performance regression: tsx probes far more filesystem paths than necessary when resolving imports, especially under the synchronous module hooks used on Node >= 24.11.1 (the path #809 bisected to in `1d7e528`).

The regression occurs because `mapTsExtensions` appends extension guesses even when the specifier has an explicit extension, leading to wasted time probing unlikely paths. For example:
- `import './config.ts'` → tries `config.ts.ts`, `config.ts.tsx`, `config.ts.jsx`, `config.ts.js`, `config.ts.json` before falling through to the real file.
- `import './foo.js'` → in addition to the desired `.js`→`.ts` check, also tries `foo.js.ts`, `foo.js.tsx`, etc.

Under the sync `registerHooks` resolver, each missing candidate further makes Node run a full CommonJS-style probe (`+.js/.json/.node/.ts/.tsx/.jsx` and `/index.*`). So a project with a single `import './leaf.ts'` issues 87 `stat` calls against its own files: 12 after this change, and 33 on 4.21.0. Projects that import with explicit extensions are particularly affected, e.g. [the repository](https://github.com/survev/survev/) linked to #809.

I had also previously noticed this [while investigating issue #640](https://github.com/privatenumber/tsx/issues/640#issuecomment-2472806032). 

## Fix

In `mapTsExtensions`:
- **Explicit TS extension** (`.ts/.tsx/.cts/.mts`) → resolve as-is, no appended guesses.
- **Swappable JS extension** (`.js/.jsx/.cjs/.mjs`) → return the swap list only; don't also append to the full path.
- **Extensionless / unknown** → guess as before (dependency `.js`-before-`.ts` ordering preserved).

## Performance

While investigating, performance was measured using two synthetic projects - one a single file import, the other a 400-module project. The counts in the following table are stat-family syscalls (`statx`/`newfstatat`/`stat`/`lstat`) issued against the project's own files, on Node 24.16.0 (Node/esbuild startup excluded).

| scenario | before (master) | after | old 4.21.0 |
|---|---|---|---|
| 400-module project, explicit `.ts` imports | 30,811 | 811 | 10,807 |
| single `import './leaf.ts'` | 87 | 12 | 33 |

### DIFF
--- a/src/utils/map-ts-extensions.ts
+++ b/src/utils/map-ts-extensions.ts
@@ -1,5 +1,10 @@
 import path from 'node:path';
-import { isFilePath, fileUrlPrefix, nodeModulesPath } from './path-utils.js';
+import {
+	isFilePath,
+	fileUrlPrefix,
+	nodeModulesPath,
+	tsExtensionsPattern,
+} from './path-utils.js';
 
 const implicitJsExtensions = ['.js', '.json'];
 const implicitTsExtensions = ['.ts', '.tsx', '.jsx'];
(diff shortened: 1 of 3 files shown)


wrose504:
Fixed by commit 257bbbb.

PR #812, opened 2026-06-28
fix: don't extend allowJs resolution into node_modules (ESM + CommonJS)

wrose504:
## Overview

With `allowJs`, tsx will resolve `.ts`/`.tsx` ahead of a requested `.js` specifier. This is usually the right heuristic for local code. In contrast, packages installed from npm or otherwise present in `node_modules` should be expected to have been transpiled already, and there's no guarantee that a `value.ts` alongside `value.js` will necessarily be transpiled correctly by tsx. Nevertheless, tsx picks the sibling `value.ts` over the dependency's own `value.js`.

This is the class of bug behind #640. It was originally diagnosed (in 2024, on tsx 4.16.3) as libphonenumber-js's internal `require('../metadata.min.json')` resolving to a sibling `metadata.min.json.js` (an ES module), so the metadata arrived as `{ default: … }` and validation threw `Got an object of shape: { default }`. The minimal reproduction below demonstrates the same principle deterministically. The real point is that tsx should not be second-guessing a dependency's own resolution process.

## Issue Reproduction

Minimal reproduction — a dependency whose entry imports `./value`, shipping both its real `value.js` and a stray `value.ts`:

```
node_modules/dep/index.js   // require('./value')  (or  export … from './value.js')
node_modules/dep/value.js   // the file the package actually ships
node_modules/dep/value.ts   // an unrelated sibling
```

With `allowJs`, tsx loads `value.ts`; it should load `value.js`

## Fix

Restrict `allowJs`'s TypeScript-first resolution to local files in the ESM hook (`resolveBase`/`resolveBaseSync`) and the CommonJS resolver (`createResolveFilename`) by excluding parents under `node_modules`. The CommonJS resolver already applies this exclusion to tsconfig path aliases; this brings extension resolution in line.

## Timeline

- `allowJs` TypeScript resolution shipped in **v4.7.3** (#535, `081853e`).
- The dependency-resolution rework in **v4.16.3** (`3df00f4`, "prioritize requested path in dependencies") extended it through the **CommonJS** resolver, introducing the regression #640 reports.
- Issue persists in both resolvers through the latest release (**v4.22.4**).

### DIFF
--- a/src/cjs/api/module-resolve-filename/index.ts
+++ b/src/cjs/api/module-resolve-filename/index.ts
@@ -94,7 +94,12 @@ export const createResolveFilename = (
 			// If parent is a TS file
 			|| (parent?.filename && tsExtensionsPattern.test(parent.filename)),
 		),
-		tsconfig?.config.compilerOptions?.allowJs ?? false,
+		// `allowJs` applies TypeScript-first resolution to local files only.
+		// Imports from inside node_modules are dependencies: their requested
+		// path should win (published packages ship runnable .js), so don't
+		// probe TypeScript extensions for them.
+		(tsconfig?.config.compilerOptions?.allowJs ?? false)
+		&& !parent?.filename?.includes(nodeModulesPath),
 	);
 
 	nextResolveSimple = createImplicitResolver(nextResolveSimple);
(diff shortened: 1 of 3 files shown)


privatenumber:
Thanks for the clear dependency-boundary analysis and regression coverage for #640. The issue is now fixed in [tsx@4.23.6](https://www.npmjs.com/package/tsx/v/4.23.6), which preserves Node resolution for dependencies under `allowJs` in both ESM and CommonJS paths.

The release includes regression coverage for a dependency requesting `metadata.min.json` beside a `.json.js` wrapper, plus validation against the original `libphonenumber-js` reproduction. This PR now conflicts with `master`, so I am closing it as superseded by the released implementation.

wrose504:
Thank you!

PR #815, opened 2026-07-03
chore: require cross-spawn 7.0.6

sonukapoor:
## Problem

The source manifest still permits `cross-spawn` 7.0.3, which is affected by [GHSA-3xgq-45jj-v275](https://github.com/advisories/GHSA-3xgq-45jj-v275). Although the current lockfile already selects patched 7.0.6, the declared range allows vulnerable versions and keeps dependency audits concerned about future source installs.

## Changes

Raise the minimum `cross-spawn` version to 7.0.6. Because latest master already resolves 7.0.6 for every consumer, this changes only the manifest range and lockfile importer specifier; the installed package graph and bundled runtime code remain unchanged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,7 @@
 		"chokidar": "^3.6.0",
 		"clean-pkg-json": "^1.2.0",
 		"cleye": "^1.3.2",
-		"cross-spawn": "^7.0.3",
+		"cross-spawn": "^7.0.6",
 		"es-module-lexer": "^1.5.4",
 		"execa": "^8.0.1",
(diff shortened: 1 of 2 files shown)


privatenumber:
Updated to 7.0.6 and `chore:` because latest master already bundles 7.0.6; this only raises the declared minimum and does not change runtime output.

I also regenerated the change from latest master and confirmed the original PR contained only the expected manifest and lockfile updates. Thanks!

privatenumber:
This issue is now resolved in [v4.23.2](https://github.com/privatenumber/tsx/releases/tag/v4.23.2).

PR #816, opened 2026-07-08
fix: ignore pnpm node_modules dependencies in watch mode

### LINKED ISSUE #810: watch: node_modules dependencies are watched in a pnpm workspace (contrary to docs)
### Acknowledgements

### Minimal reproduction URL

https://github.com/SebT/tsx-emfile-repro

`tsx watch` restarts when a file **inside `node_modules`** changes, although the docs say `node_modules` is ignored by default. In a pnpm workspace every imported dependency gets watched.

Repro (`./reproduce.sh`): a subpackage imports `nanoid` and idles; touching nanoid's file under `node_modules` makes tsx restart, in its own words:
`[tsx] change in ./../../node_modules/.pnpm/nanoid@5.1.15/node_modules/nanoid/index.js Restarting...`
Also reproduced on a clean Linux runner (workflow in the repo).

**Why pnpm:** a dep's real files live at `node_modules/.pnpm/<pkg>@<ver>/node_modules/<pkg>`, resolved from a subpackage as `../../node_modules/...`. The default `node_modules` filter doesn't match that parent-relative path.

**`--exclude`:** `--exclude '**/node_modules/**'` is a silent no-op; only `--exclude '../../node_modules/**'` (or an absolute path) works — tsx seems to match `--exclude` against absolute realpaths.

**Impact:** across ~10 services each tsx holds thousands of `node_modules` watches (≈50 under `ts-node-dev`); on macOS this exhausts the system file table (`kern.maxfiles`) and an unrelated `fs.watch` dies with `EMFILE: too many open files, watch`.

**Expected:** the default `node_modules` exclusion should cover pnpm's `.pnpm` realpaths, and/or `--exclude '**/node_modules/**'` should match them.

cyphercodes:
## Summary

- Skip runtime dependency watch entries that resolve inside default third-party directories (`node_modules`, `bower_components`, `vendor`).
- Add the same predicate to chokidar's ignored paths so pnpm realpaths outside the current workspace package stay ignored.
- Add a regression test for a pnpm-style symlinked dependency whose real files live under `../../node_modules/.pnpm/...`.

Fixes #810

## Tests

- `pnpm type-check`
- `pnpm lint` (passes with existing warnings)
- `node ./dist/cli.mjs /tmp/oss-pr-pipeline/watch-runner.mts` (targeted watch suite; 12 passed)
- `pnpm test` attempted: the watch suite (including this regression) passed, but the full run failed in unrelated existing/environment-sensitive smoke named-export checks and PTY Ctrl+C signal tests in this container.

### DIFF
--- a/src/watch/index.ts
+++ b/src/watch/index.ts
@@ -17,6 +17,18 @@ import {
 	log,
 } from './utils.js';
 
+const ignoredDependencyDirectories = new Set([
+	'node_modules',
+	'bower_components',
+	'vendor',
+]);
+
+const isInIgnoredDependencyDirectory = (
+	filePath: string,
+) => filePath.split(/[\\/]+/).some(
+	pathSegment => ignoredDependencyDirectories.has(pathSegment),
+);
+
 const flags = {
 	noCache: {
 		type: Boolean,
@@ -97,7 +109,10 @@ export const watchCommand = command({
 					: data.path
 			);
 
-			if (path.isAbsolute(dependencyPath)) {
+			if (
+				path.isAbsolute(dependencyPath)
+				&& !isInIgnoredDependencyDirectory(dependencyPath)
+			) {
 				watcher.add(dependencyPath);
 			}
 		}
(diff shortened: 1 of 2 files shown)


jamalkamaladdin (commented):
#830 takes the same approach for #810 and applies cleanly, while this branch now conflicts with master.

The part #830 leaves out is passing the predicate to `watchOptions.ignored`, which is what covers pnpm realpaths resolving outside the workspace. If that case still matters, it could be carried over there.

PR #823, opened 2026-07-15
fix: preserve listenerCount overload semantics

### LINKED ISSUE #822: process.listenerCount(event, listener) reports incorrect counts under CLI
## Description

The CLI installs hidden `SIGINT` and `SIGTERM` relay listeners and wraps `process.listenerCount()` so those internal listeners are not visible. The wrapper currently subtracts one from every count for those signals, including the `listenerCount(eventName, listener)` overload and cases where the hidden listener has already been removed.

## Reproduction

```js
const handler = () => {};
process.on('SIGINT', handler);

setTimeout(() => {
	console.log(process.listenerCount('SIGINT', handler));
	process.removeAllListeners('SIGINT');
	console.log(process.listenerCount('SIGINT'));
}, 100);
```

With Node v24.18.0:

```console
$ node repro.js
1
0
```

With the current `tsx` master (`5828ffe`):

```console
$ tsx repro.js
0
-1
```

## Expected behavior

Specific user-listener counts should match Node, aggregate counts should never become negative, and the internal relay listener should remain hidden while it is installed.

ShiroKSH:
Fixes #822

## Summary

- Preserve Node's listener-specific `process.listenerCount(eventName, listener)` semantics while keeping tsx's internal signal relay hidden.
- Subtract only relay listeners that are still registered, preventing negative counts after `removeAllListeners()`.
- Add a deterministic regression test covering specific listeners, invalid listener values, and relay removal.

## Root cause

The preflight wrapper unconditionally subtracted one for `SIGINT` and `SIGTERM`. That also changed listener-specific queries and assumed the hidden relay still existed.

## Validation

- Targeted signal tests on Node 24.15.0: 2 passed
- Targeted signal tests on Node 18.20.8: 2 passed
- Manual compatibility check on Node 18.0.0
- `pnpm type-check`
- Changed-file lint: 0 errors

### DIFF
--- a/src/preflight.cts
+++ b/src/preflight.cts
@@ -37,8 +37,30 @@ const bindHiddenSignalsHandler = (
 
 	process.listenerCount = function (eventName) {
 		let count = Reflect.apply(listenerCount, this, arguments);
-		if (signals.includes(eventName as RelaySignals)) {
-			count -= 1;
+		const hiddenHandler = hiddenHandlers.get(eventName as RelaySignals);
+		const listener = arguments[1] as unknown;
+		const isAggregateCount = (
+			listenerCount.length === 1
+			|| arguments.length < 2
+			|| listener === undefined
+			|| listener === null
+		);
+		if (
+			hiddenHandler
+			&& (
+				isAggregateCount
+				|| listener === hiddenHandler
+			)
+		) {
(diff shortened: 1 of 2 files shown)


privatenumber:
This issue is now resolved in [v4.23.3](https://github.com/privatenumber/tsx/releases/tag/v4.23.3).

PR #827, opened 2026-07-26
fix(cli): allow async process.once() signal handlers to finish

### LINKED ISSUE #826: SIGINT / SIGTERM kill process too early, preventing async operations during shutdown (tsx v4.23.1)
### Acknowledgements

### Minimal reproduction URL

https://github.com/rameshvarun/tsx-graceful-shutdown-repro

Under the tsx CLI, SIGINT terminates the process before an async SIGINT handler finishes. `node` and `node --import tsx` let it run to completion.

I couldn't repro this on Stackblitz, possibly due to differences in signal handling on WebContainer vs. a real UNIX system. I was able to repro on my Mac and in Docker.

- This is not the same as https://github.com/privatenumber/tsx/issues/708 as `npx` is not involved.
- This is not the same as https://github.com/privatenumber/tsx/issues/95 since it can be reproduced on the latest TSX version v4.23.1

One line repro in Docker command: `docker run --rm --init node:24.16.0 bash -lc 'git clone --depth 1 https://github.com/rameshvarun/tsx-graceful-shutdown-repro.git /app && cd /app && npm ci && npm start'`

- [x] I'm interested in working on this issue

rameshvarun:
Fixes #826

## Problem

When an application run through the tsx CLI handles `SIGINT` or `SIGTERM` with `process.once()`, tsx can exit before asynchronous cleanup finishes.

tsx installs a hidden listener to relay signals to its parent process and emulate Node's default exit when no application listeners remain. That listener could run after an application's one-time listener. Node removes a one-time listener before invoking it, so tsx then saw no remaining application listeners and called `process.exit()`.

An application could start closing a database connection, then exit before its cleanup completes. The same handler completes when run with `node` or `node --import tsx`.

## Changes

tsx now runs its hidden relay listener before normal application listeners. It observes a one-time listener before Node removes it, allowing the handler to complete asynchronous cleanup and control the exit code.

The regression sends `SIGINT` to an asynchronous `process.once()` handler and confirms that cleanup finishes before exit. The test fails on the pre-fix implementation with exit code `130` and passes with exit code `42` after cleanup completes.

This covers ordinary `process.once()` registration. A handler added later with `process.prependOnceListener()` can still intentionally take precedence over the relay.

### DIFF
--- a/src/preflight.cts
+++ b/src/preflight.cts
@@ -26,9 +26,7 @@ const bindHiddenSignalsHandler = (
 			}
 		};
 
-		/**
-		 * Prepend so this runs before user handlers.
-		 */
+		// Prepend so this runs before user handlers.
 		process.prependListener(signal, hiddenHandler);
 		hiddenHandlers.set(signal, hiddenHandler);
 	}
--- a/src/preflight.cts
+++ b/src/preflight.cts
@@ -26,7 +26,10 @@ const bindHiddenSignalsHandler = (
 			}
 		};
 
-		process.on(signal, hiddenHandler);
+		/**
+		 * Prepend so this runs before user handlers.
+		 */
+		process.prependListener(signal, hiddenHandler);
 		hiddenHandlers.set(signal, hiddenHandler);
 	}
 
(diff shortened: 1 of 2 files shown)


privatenumber:
This issue is now resolved in [v4.23.4](https://github.com/privatenumber/tsx/releases/tag/v4.23.4).

Issue #828, opened 2026-08-05
`import.meta` is left unshimmed by the CJS transform when comments or newlines split the tokens


AtofStryker:
### Acknowledgements

- [x] I read the documentation and searched existing issues to avoid duplicates
- [x] I understand this is a **bug tracker** and anything other than a proven bug will be closed
- [x] I understand this is a free project and relies on community contributions
- [x] I read and understood the [Contribution guide](https://github.com/privatenumber/tsx/blob/master/CONTRIBUTING.md)

### Minimal reproduction URL

https://stackblitz.com/edit/tsx-reproduction?file=package.json,entry.cjs,deps.js

### Problem & expected behavior (under 200 words)

Requiring an ESM file through the CJS hook crashes when a comment or newline sits between the `import` and `.meta` tokens:

```
TypeError [ERR_INVALID_ARG_VALUE]: The argument 'filename' must be a file URL object,
file URL string, or absolute path string. Received undefined
    at createRequire (node:internal/modules/cjs/loader:2025:11)
```

The failing form is `createRequire(import /*::(_)*/.meta.url)` — valid syntax, and plain `node entry.cjs` (native `require(esm)`, Node 20.19+) loads it fine. Under tsx, `import.meta` lowers to an empty object and `.url` is `undefined`.

Cause: `transformSync` in `src/utils/transform/index.ts` only passes the `import.meta` define to esbuild when `code.includes('import.meta')`. Comments or newlines between the tokens defeat the textual search. esbuild's define itself handles those forms correctly once configured.

This is hit in the wild: `@babel/helper-define-polyfill-provider@1.x` ships exactly that form (a preserved Flow type-comment — [source at the 1.0.0 tag](https://github.com/babel/babel-polyfills/blob/%40babel/helper-define-polyfill-provider%401.0.0/packages/babel-helper-define-polyfill-provider/src/node/dependencies.ts#L5)), so any CJS code that pulls in `@babel/preset-env` 8 under the hook crashes — e.g. `@angular-devkit/build-angular` 22.1, found via cypress-io/cypress#34461.

Reproduced on 4.22.4 through 4.23.8.

### Bugs are expected to be fixed by those affected by it

- [x] I'm interested in working on this issue

### Compensating engineering work will speed up resolution and support the project

- [ ] I'm willing to offer $10 for financial support

privatenumber:
This issue is now resolved in [v4.23.12](https://github.com/privatenumber/tsx/releases/tag/v4.23.12).

[...]

PR #829, opened 2026-08-05
fix: shim `import.meta` when tokens are split by comments or newlines

AtofStryker:
## Problem

[#828](https://github.com/privatenumber/tsx/issues/828) found that valid `import.meta` expressions can include whitespace or comments between `import` and `.meta`. The previous text match missed those forms, leaving CommonJS-transformed modules without the metadata Node would normally provide.

## Changes

tsx now uses its existing `es-module-lexer` integration to identify `import.meta` in eligible CommonJS transforms. The regression test covers comment- and newline-separated forms.

### DIFF
--- a/src/utils/transform/index.ts
+++ b/src/utils/transform/index.ts
@@ -8,6 +8,7 @@ import {
 	type TransformFailure,
 } from 'esbuild';
 import { sha1 } from '../sha1.js';
+import { parseEsm } from '../es-module-lexer.js';
 import { importMetaPathProperties, isFeatureSupported } from '../node-features.js';
 import {
 	version as transformDynamicImportVersion,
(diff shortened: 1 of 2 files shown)


privatenumber:
This regexp reimplements JavaScript tokenization. It has a known quadratic case and misses valid `//` terminators such as `\r`, U+2028, and U+2029, so it still does not cover the full grammar.

A coarse `code.includes('import')` prefilter can leave syntax detection to esbuild while preserving the fast path. I’ll follow up on this branch with that simplification.

privatenumber:
Follow-up: tsx already uses `es-module-lexer`, which identifies `import.meta` entries with `d === -2`, including comment- and newline-separated forms. I’ve switched this branch to that exact lexer result, superseding the coarse-prefilter approach above.

privatenumber:
This issue is now resolved in [v4.23.12](https://github.com/privatenumber/tsx/releases/tag/v4.23.12).

PR #830, opened 2026-08-11
fix: ignore node_modules dependencies in watch mode

### LINKED ISSUE #810: watch: node_modules dependencies are watched in a pnpm workspace (contrary to docs)

MykalMachon:
## Summary of changes

Skip runtime dependencies inside ignored third-party directories before adding them to the watcher. This stops tsx from exhausting file descriptors with redundant watchers when `node_modules` packages are symlinked, similar to how pnpm resolves cached dependencies.

Adds regression coverage for symlinked `node_modules`. Follows the approach proposed in #816, updated for the current watch implementation.

Fixes #810

### DIFF
--- a/src/watch/index.ts
+++ b/src/watch/index.ts
@@ -19,6 +19,15 @@ import {
 	log,
 } from './utils.js';
 
+const ignoredDependencyDirectories = new Set([
+	'node_modules',
+	'bower_components',
+	'vendor',
+]);
+const isInIgnoredDependencyDirectory = (filePath: string) => filePath
+	.split(/[\\/]+/)
+	.some(pathSegment => ignoredDependencyDirectories.has(pathSegment));
+
 const flags = {
 	noCache: {
 		type: Boolean,
@@ -163,7 +172,10 @@ export const watchCommand = command({
 					: data.path
 			);
 
-			if (path.isAbsolute(dependencyPath)) {
+			if (
+				path.isAbsolute(dependencyPath)
+				&& !isInIgnoredDependencyDirectory(dependencyPath)
+			) {
 				if (exactNegatedIncludes.has(getPathKey(dependencyPath))) {
 					if (!dependencyOverrideWatcher) {
 						dependencyOverrideWatcher = new FSWatcher({
--- a/tests/specs/watch.ts
+++ b/tests/specs/watch.ts
@@ -592,6 +592,48 @@ export const watch = ({ tsx, path: nodePath }: NodeApis) => describe('watch', as
 
 			await tsxProcess;
 		}, 10_000);
+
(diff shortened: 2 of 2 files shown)


jamalkamaladdin (commented):
This overlaps with #816. Both add the same `ignoredDependencyDirectories` set and the same path predicate to `src/watch/index.ts`, and both close #810. #816 currently conflicts with master, so this branch looks like the one to keep, but only one of them should land.

Two notes on this version:

1. `node_modules`, `bower_components` and `vendor` are now listed twice: in the new set, and in the existing glob in `watchOptions.ignored` (`'**/{node_modules,bower_components,vendor}/**'`, `src/watch/index.ts:302`). If one list changes later, the other drifts silently. Deriving one from the other keeps them in sync.

2. The new condition runs before the `exactNegatedIncludes` branch, so a runtime dependency inside `node_modules` can no longer reach `dependencyOverrideWatcher`. With `--include='!./node_modules/pkg/index.js'` that path is now dropped instead of being watched by the override watcher. Is that intended? The existing test at `tests/specs/watch.ts:150` uses a file outside `node_modules`, so it passes either way.

One thing #816 does and this does not: it also passes the predicate to `watchOptions.ignored`, which is what catches pnpm realpaths resolving outside the workspace. If that case is part of #810, it may be worth carrying over here.

PR #835, opened 2026-08-28
fix(cache): bound shared transform cache memory

### LINKED ISSUE #834: Shared transform cache size can exhaust heap before entrypoint execution
### Acknowledgements

- [x] I read the documentation and searched existing issues to avoid duplicates.
- [x] I understand this is a bug tracker and anything other than a proven bug will be closed.
- [x] I understand this is a free project and relies on community contributions.
- [x] I read and understood the contribution guide.

### Minimal reproduction URL

https://github.com/steipete/tsx/blob/771864e6fcfa6e35af52acb0dc3326df7d2ef1d5/tests/specs/transform-cache.ts#L18-L52

This test-only commit retains the original cache implementation. After checking out the commit, run the focused spec in an owned temporary namespace:

```bash
TMPDIR="$(mktemp -d)" bash -ceu '
  export TMP="$TMPDIR" TEMP="$TMPDIR" TSX_DISABLE_CACHE=1
  unset NODE_OPTIONS NODE_PATH
  pnpm install --frozen-lockfile --ignore-scripts
  pnpm build
  node dist/cli.mjs --eval "import(\"./tests/specs/transform-cache.ts\").then(module => module.transformCacheSpec())"
'
```

The outer tooling skips disk caching; the regression explicitly reenables it for each owned child. On Node 24.15.0, the empty 32 MiB control passes and the 150,000-entry case aborts with V8 heap exhaustion. The original eight cache tests pass. Hosted before/after CI evidence will follow.

The reproduction uses only synthetic files in owned temporary directories. It does not inspect or delete an existing shared cache.

With tsx 4.23.12 and Node 26.7.0, `const answer: number = 42; console.log(answer)` works with an empty cache but exhausts V8 old space before executing when the cache contains 713,453 unrelated fresh entries. There are no application imports or inherited `NODE_OPTIONS`/`NODE_PATH`.

On Linux/arm64, the real loader fails at 128 MiB for `.ts`, `.mts`, and `.cts`; `.ts` and `.cts` also fail at 256 MiB. Empty, 100,000-entry, and 300,000-entry controls pass. Separately, 100,000 expired entries can let an `.mts` program print and then cause an OOM during cleanup at 128 MiB.

Expected: a tiny program's cache memory should not scale with unrelated transforms from other projects.

`getDiskCacheIndex()` synchronously materializes the entire directory into a Map and entry array before scheduling expiration. Expiration queues all unlink promises and repeatedly scans/splices that array. Separate ESM/CJS bundles can each build an index in one thread.

I am preparing an owner-level repair that retains disk-cache reuse while bounding lookup and expiration memory.

- [x] I'm interested in working on this issue.

steipete:
## Problem

tsx stores transformed modules in a shared disk cache so later processes can reuse them. The first cache lookup currently reads every filename into an in-memory index, including entries unrelated to the requested module.

As the shared cache grows, a small TypeScript program can exhaust its heap before its entrypoint runs. A cache with 713,453 unrelated entries reproduced the failure at a 128 MiB heap limit, and expiration of 100,000 entries could also exhaust memory after execution. Fixes #834.

## Changes

Cache reads now probe only the requested SHA-1 key in the current and previous seven timestamp buckets. A lookup therefore uses constant memory regardless of how many unrelated files share the cache. Valid hits remain cached in the process, corrupt current entries can fall back to an older live entry, and reads do not create or maintain cache directories.

Writes still create the cache and schedule best-effort maintenance. Expiration now streams directory entries and processes them in batches of 64, which bounds outstanding unlink promises without retaining a directory-wide index. The cache filename format and seven-bucket retention policy remain unchanged.

The regression guard asserts that lookup never calls `readdirSync()`. It fails against the original implementation with one directory enumeration and passes with the bounded lookup. The focused suite also covers live-window lookup, corrupt entries, expiration boundaries, malformed filenames, and concurrent cache writers.


privatenumber:
This issue is now resolved in [v4.23.13](https://github.com/privatenumber/tsx/releases/tag/v4.23.13).

steipete:
sweet, thank you!

steipete:
I already sponsor you through the OpenClaw Foundation, and just increased level from bronze to silver. Thanks again!

privatenumber:
Thanks @steipete ! 🙌❤️

